PR #9, opened 2013-10-07
Coordinate Conversion Snippet

keflavich:
A tutorial on coordinate declaration and conversion and access, copied directly from the astropy paper.


hamogu:
Any progress on this since April 2014? My feeling is that the astropy docs are good enough in this area that we might not need this as a separate tutorial (so this PR could be closed), but if it's updated to the current coordinates syntax, it might be worth having.

eteq:
I'm inferring from the conversations above that this is probably not worth resurrecting (and perhaps was superceded anyway by #97 ?), so I'm closing this.  But if you want to bring it back at some point, feel free to do so, @keflavich

keflavich:
@eteq yes, closing was correct, thanks.

PR #49, opened 2014-02-20
Add a Units tutorial

keflavich:
I think this is the first thing I want to teach folks here; it's super easy to use and incredibly useful on a daily basis.

I'd appreciate feedback - @kelle, a checklist for important things to add would be great.  Probably the starting place should be the astropy docs.  Right now the units demo isn't much different from the docs.  

@adrn, I think you had an equivalent notebook at some point?  Perhaps we should merge these?  Or just use yours?

### DIFF
--- a/tutorials/Units/Units.ipynb
+++ b/tutorials/Units/Units.ipynb
@@ -1,0 +1,296 @@
+{
+ "metadata": {
+  "name": "",
+  "signature": "sha256:64d8cc1db960c9a9de019b9add01ab17b5e269058ee1ac21f6d4d9745d87c179"
+ },
+ "nbformat": 3,
+ "nbformat_minor": 0,
+ "worksheets": [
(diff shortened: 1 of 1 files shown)


hamogu:
Any progress on this since April 2014? Or is this not needed any longer and the PR can be closed?

eteq:
@hamogu raised a good question here.  @keflavich, what do you think?  The obvious comparison is #51, which is a pretty thorough treatment of quantities.  Do you think this has more that should still be included, or should we just close this one?

hamogu:
Quantities are also used as an example in other notebooks, e.g. see "Units and constants in astropy" in http://www.astropy.org/astropy-tutorials/UVES.html

keflavich:
Let's drop it, then, and I'll add to the other tutorials as needed.

Issue #67, opened 2014-07-07
Bug with RA increasing to the right


astrofrog:
This plot in Quantities is wrong:

![unknown](https://cloud.githubusercontent.com/assets/314716/3501932/92dd56f0-061b-11e4-9a56-2d866db2e906.png)

The RA should increase to the left! Same goes with the plot-catalog tutorial:

![unknown-1](https://cloud.githubusercontent.com/assets/314716/3501940/a438f580-061b-11e4-8281-a6393bf9c85a.png)

adrn:
This was addressed with #70

PR #70, opened 2014-11-05
Fix/flip longitude where possible. Explain where it is not.

jfoster17:
This is in response to #67. Unfortunately there is no easy fix for the inability to flip the RA/longitude axis when using matplotlib projections. The only current fix seems to require the Basemap Toolkit, which we don't want to require.


jfoster17:
I'm keeping "highly non-trivial", as it expresses my frustration.

PR #71, opened 2014-11-06
Increase contrast and switch default font for tutorials

### LINKED ISSUE #66: Change plotting font?
I was wondering whether it would make sense to switch the font to a 'modern' font? While the current 'retro' look works in some cases, I'm a little worried that it gives a PGPLOT look to some of the plots which might make people think that the plotting package is old. Matplotlib can make really nice modern-looking plots, so it'd be a shame to miss out on that. Using Arial (with bold axis labels) works nicely for example.

jfoster17:
This PR increases the contrast in the default matplotlibrc used in tutorials, and switches back to a more modern-looking font (resolves #66). Of course such things are subjective, but I feel strongly that the current style emphasizes aesthetics (nice, soft-looking) over legibility, particularly for anyone with less-than-perfect vision.

Old style
old_plot
New style
new_plot

### DIFF
--- a/templates/matplotlibrc
+++ b/templates/matplotlibrc
@@ -202,6 +202,7 @@ axes.edgecolor      : black   # axes edge color
 axes.linewidth      : 1        # edge linewidth
 axes.grid           : True     # display grid or not
 axes.titlesize      : x-large  # fontsize of the axes title
+axes.labelweight     : bold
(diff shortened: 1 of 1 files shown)


jfoster17:
Finally got around to checking this -- all the plots look okay when I run the conversion, so I'm going to merge this now.

PR #75, opened 2015-02-26
remove unneeded np.where in FITS-tables

eteq:
While showing some of the tutorials to some people, I noticed in the FITS-tables tutorial that there's a use of `np.where` that's unnecessary.  That is, a boolean mask will work just fine.

I find this is a common mistake people make transitioning from IDL, so I think it's much better if we avoid using ``np.where`` wherever possible.  So this change just removes the ``np.where`` call.

Is that ok with you @adrn and @eblur (I think you wrote this originally, right @eblur)?

### DIFF
--- a/tutorials/FITS-tables/FITS-tables.ipynb
+++ b/tutorials/FITS-tables/FITS-tables.ipynb
@@ -3,12 +3,12 @@
   "astropy-tutorials": {
    "author": "Lia R. Corrales <lia@astro.columbia.edu>",
    "date": "January 2014",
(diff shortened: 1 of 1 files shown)


astrofrog:
:+1:

eblur:
sounds good to me.

On Thu, Feb 26, 2015 at 4:58 AM, Thomas Robitaille <notifications@github.com

adrn:
:+1:

jfoster17:
P.S. I ran deploy to update the html version as well

Issue #64, opened 2014-06-23
Add a script for checking that requirements are met for running tutorial


taldcroft:
For the SciPy tutorials they asked as to prepare a script to verify that the required packages were available to run the tutorial.  The result was:
https://github.com/astropy/astropy4scipy2014/blob/master/check_env.py

This was thrown together very quickly, but something along these lines might be a good idea for the Astropy tutorials as well.  It might make sense to have a `check_env` script for each tutorial since the requirements for different tutorials might be different.

Note that we uncovered an apparent bug in `distutils.version.LooseVersion` in that it fails to support comparison of versions that contain strings (like `dev`) in Python 3.  Hence the current `check_env.py` script just does a string comparison of versions, but this would need to be fixed.

amanXCalibre:
Hey @taldcroft  does this pull request fix this issue ?
https://github.com/astropy/astropy-tutorials/pull/83

adrn:
This can now be done with setuptools `pkg_resources` with the one-liner:
```bash
python -c "import pkg_resources; pkg_resources.require(open('pip-requirements.txt', mode='r')); print('Your environment is all set!')"
```
Maybe we should add this to our readme?

Issue #50, opened 2014-03-09
Add Disqus comments at the bottom of each tutorial


astrofrog:
It would be reasonably easy to update the ``setup.py`` to inject a Disqus comment thread at the bottom of each notebook. What do you think?

eteq:
Dine in #77

PR #83, opened 2015-04-08
Fixing Issue #64

x-calibre:
Hey @taldcroft Regarding Issue 64 ...does this solution fix this?
I modified each check_env.py to check for packages that I thought were required , some packages might be extra or missing...Do tell me if you find something like this
Another potential issue is the oldest version of packages supported and the most compatible ones ( like in numpy is anything below 1.6.0 supported , if yes, till which version) maybe be absent or wrongly coded.

### DIFF
--- a/check_env.py
+++ b/check_env.py
@@ -24,23 +24,20 @@ import json
 try:
     import astropy
 except ImportError:
-    import logging as logger
-    logger.error("'Astropy' package is missing and is a core dependency to run this script.\
-    \nFor more information visit http://www.astropy.org/")
-    quit()
+    raise ImportError("'Astropy' package is missing and is a core dependency "
+                      "to run this script. For more information visit "
+                      "http://www.astropy.org/")
     #Astropy is necessary to run tutorials of astropy. If not present, stop immediately
-
-
-from astropy import log as logger
 
 try:
     import six
 except ImportError:
-    logger.error("Python library 'six' is missing and is required to run this script.\
-    \nFor more information visit https://pypi.python.org/pypi/six")
-    quit()
+    raise ImportError("'Six' package is missing and is a core dependency "
+                      "to run this script. For more information visit "
+                      "https://pypi.python.org/pypi/six/")
     #six is required for checking the versions. If not present, stop immediately
 
+from astropy import log as logger
 
 def check_environment(tutorial=None):
     error = False
--- a/check_env.py
+++ b/check_env.py
@@ -10,47 +10,57 @@ import six
 
 from astropy import log as logger
 
+
 def check_environment(tutorial=None):
-    error=False
-    enter=False
-    warnings=False
+    error = False
+    enter = False
+    warnings = False
     logger.info("Running Environment Tests...")
     _orig_path = os.getcwd()
-    tutorials_base = os.path.join(_orig_path,'tutorials')
-    for tutorial_name in os.listdir(tutorials_base) :
-        if (tutorial_name==tutorial or tutorial==None):
-            enter=True
+    tutorials_base = os.path.join(_orig_path, 'tutorials')
+    for tutorial_name in os.listdir(tutorials_base):
+        if (tutorial_name == tutorial or tutorial is None):
+            enter = True
             tutorial_path = os.path.join(tutorials_base, tutorial_name)
             try:
-                with open((tutorial_path+"/requirements.json")) as data_file:
-                    data=json.load(data_file)
+                with open((tutorial_path + "/requirements.json")) as data_file:
+                    data = json.load(data_file)
                     for pkgname in data:
                         for subinfo in data[pkgname]:
-                            if subinfo=='min_version':
-                                if(not minversion(pkgname,data[pkgname][subinfo])):
-                                    logger.error("Package "+pkgname+" doesn't satisfy requirements of Tutorial: "+tutorial_name)
-                                    error=True
-                            elif subinfo=='pref_version':
-                                if(not minversion(pkgname,data[pkgname][subinfo])):
-                                    logger.warning("Please upgrade Package "+pkgname+" to version "+data[pkgname][subinfo]+" for Tutorial: "+tutorial_name)
-                                    warnings=True
+                            if subinfo == 'min_version':
+                                if(not minversion(pkgname, data[pkgname][subinfo])):
+                                    logger.error(
+                                        "Package " +
+                                        pkgname +
+                                        " doesn't satisfy requirements of Tutorial: " +
+                                        tutorial_name)
+                                    error = True
+                            elif subinfo == 'pref_version':
+                                if(not minversion(pkgname, data[pkgname][subinfo])):
+                                    logger.warning(
+                                        "Please upgrade Package " +
+                                        pkgname +
+                                        " to version " +
+                                        data[pkgname][subinfo] +
+                                        " for Tutorial: " +
+                                        tutorial_name)
+                                    warnings = True
             except IOError:
-                logger.error("Environment Check Failed ! requirements.json not found");
+                logger.error(
+                    "Environment Check Failed ! requirements.json not found")
 
-    if enter==False:
+    if not enter:
         logger.error("Wrong tutorial name entered !")
         return
-    if warnings==True:
+    if warnings:
         logger.warning("Please resolve the above warnings soon")
-    if error==False:
+    if not error:
         logger.info("Environment Check Passed")
     else:
         logger.info("Please resolve the above errors to continue!")
 
 
-
-
-#ripped off from astropy.utils.introspection
+# ripped off from astropy.utils.introspection
 
 
 def resolve_name(name):
@@ -115,7 +125,7 @@ def resolve_name(name):
     return ret
 
 
-def minversion(module, version,inclusive=True, version_path='__version__'):
+def minversion(module, version, inclusive=True, version_path='__version__'):
     """
     Returns `True` if the specified Python module satisfies a minimum version
     requirement, and `False` if not.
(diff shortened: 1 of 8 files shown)


taldcroft:
@amanXCalibre - I think this could be handled with a single `check_env.py` module at the top level and then a `requirements.json` file in each tutorial directory.  Here's an example of such a file:
```
{
    "joe": {
        "min_version": "2.0"
    }, 
    "matplotlib": {
        "min_version": "1.1.0"
    }
}
```
Even though JSON appears to be too much overhead for such a simple requirements file, using this format allows for future expansion.  For instance having a `preferred_version` tag which would also include the warning message, etc.  Right now I don't actually think we need a `preferred_version`.  The part in the original `check_env.py` script which allowed astropy 0.3 but preferred 0.4 was a bit of a stop-gap for the astropy tutorial at SciPy where 0.4 was not yet officially released but was available for the tutorial.  At this time I don't see a strong reason to not simply require 0.4.

`check_env.py` could be imported into `prepare_deploy.py` and also be run standalone by users with something like `check_env.py Quantities`.

taldcroft:
Also, you should copy the code in the current `astropy.utils.introspection.minversion` into the new `check_env.py` and use that instead of simple string comparison (which is not reliable in general).

amanXCalibre:
@taldcroft  Did everything as you said. There is a check_env.py on the top level....it can be run separately as 'python check_env.py' or 'python check_env.py -n  {package name}'. It can also be run from 'python prepare_deploy.py check' .The tutorial name can also be sent to check_env.check_environment(tutorial_name) to test if all requirements are met within the prepare_deploy script.
I did snip off astropy.utils.introspection.minversion and added it to check_env. Will need to modify it a bit in the future to include "preferred version" case. 
All the current requirements.json files are the same (the one which you wrote in the comments earlier) and should do for now!

amanXCalibre:
Did everything as you said. There is a check_env.py on the top level....it can be run separately as 'python check_env.py' or 'python check_env.py -n {package name}'. It can also be run from 'python prepare_deploy.py check' .The tutorial name can also be sent to check_env.check_environment(tutorial_name) to test if all requirements are met within the prepare_deploy script.
I did snip off astropy.utils.introspection.minversion and added it to check_env.
All the current requirements.json files are the same (the one which you wrote in the comments earlier) and should do for now! If the user wishes to check for a preferred version , the json variable 'pref_version: version no.' should be added.

check_env.py
```diff
@@ -0,0 +1,196 @@
...
+                        for subinfo in data[pkgname]:
+                            if subinfo=='min_version':
+                                if(not minversion(pkgname,data[pkgname][subinfo])):
+                                    logger.error("Package "+pkgname+" doesn't satisfy requirements of Tutorial: "+tutorial_name)
+                                    error=True
+                            elif subinfo=='pref_version':
+                                if(not minversion(pkgname,data[pkgname][subinfo])):
+                                    logger.warning("Please upgrade Package "+pkgname+" to version "+data[pkgname][subinfo]+" for Tutorial: "+tutorial_name)
+                                    warnings=True
+            except:
```
taldcroft: Use `except IOError` to only catch the "expected" error of a missing file.

check_env.py
```diff
@@ -0,0 +1,196 @@
...
+import six
+
+from astropy import log as logger
+
+def check_environment(tutorial=None):
+    error=False
+    enter=False
+    warnings=False
+    logger.info("Running Environment Tests...")
+    _orig_path = os.getcwd()
```
taldcroft: I think you can make this a little cleaner by not changing directories but instead just construct the full path to `requirements.json` using `tutorial_path`.  Not a big deal but it's often a good idea to not use `chdir` unless you really need to do so.

check_env.py
```diff
@@ -0,0 +1,196 @@
...
+    if error==False:
+        logger.info("Environment Check Passed")
+    else:
+        logger.info("Please resolve the above errors to continue!")
+
+    os.chdir(_orig_path)
+
+
+
+#ripped off from astropy.utils.introspection
```
taldcroft: Better to be a little more formal and descriptive in comments and saying something like
```
# Functions resolve_name() and minversion() were copied from astropy.utils.introspection
```

taldcroft:
@amanXCalibre - this looks awesome!

My only two remaining comments are:

- There are a bunch of PEP8 formatting violations.  Not a huge deal but we'd like to have PEP8 compliant code.  If you just install and run `autopep8` on *your new files* then everything should be fine.
- No need for formal testing, but can you include here logs showing running the tool where the various warnings or errors get generated?

amanXCalibre:
@taldcroft  Sorry for the sloppy comments. I intended to brush them up after I have got the basic script right. I have added a few more comments (formal , this time :) ) and a small guide as to what  the script does and how to run it. The entire code now is PEP8 compliant. All previous issues you pointed out have been handled. Thanks for pointing them out,
I will upload the log soon enough.

adrn:
@taldcroft displaying my own git-norance here, but are the first two commits going to cause any trouble when #50 gets merged?

amanXCalibre:
You're right...Should I make a new pull request...with a clean slate then @adrn  ?

taldcroft:
@amanXCalibre @adrn - about the first two commits -- this won't cause any problem but they won't go away and should probably be cleaned up.  This is straightforward with rebase.  In your current branch do this: `git rebase -i xxxxxxx` where xxxxxxx is the git hash tag in the git log just *before* your first commit in this branch 623ea5c.  This will bring up an editor and you can choose to ignore the first two commits and leave all the rest as-is (pick).

BTW, for safety it's always a good idea to do something like `git branch aman-bak` first to make it easier to recover from a rebase gone bad.  Once you have your `aman` branch looking OK then force push it (the `-f` option to push).

amanXCalibre:
@taldcroft  Thank you so much for the easy-to-follow instructions ! I had read about rebasing only in theory and it was fun (and scary) to try out ! Seems it went well !

amanXCalibre:
Is this ready to merge ?

adrn:
I'm happy with this! Any comments @eteq or @taldcroft ?

check_env.py
```diff
@@ -0,0 +1,238 @@
...
+                        unicode_literals)
+# Standard library
+import os
+import json
+import inspect
+import sys
+import types
+import six
+
+from astropy import log as logger
```
taldcroft: This should be wrapped in a try / except ImportError, with a message that astropy is missing and needs to be installed.  It's probably expected that anyone running astropy tutorials would have some version of astropy installed, but for beginners you never know.  They should get a friendly message if not.  :smile:

taldcroft:
Apart from my one comment this is good to merge.

amanXCalibre:
@taldcroft  Thanks for pointing that out ! I have fixed the issue as you said. Just out of curiosity,would it not be better to do all the logging for the check_env.py script using Python's default `logging` library...This way the script can be practically independent of all libraries and the code would be more universal.

amanXCalibre:
Also @taldcroft , I ran the script on a fresh python environment and `six` wasn't present in the environment , so I  added a check for `six` also manually. New users may not have that installed as well. Hopefully all the cases are covered now !

taldcroft:
@amanXCalibre  @adrn this looks good to me now!

amanXCalibre:
Thank you so much for reviewing it again !

adrn:
Hey @x-calibre -- sorry I lost track of this PR. When reviewing again just now, I had a question and can't remember why we decided to do this: why did you copy the functions `resolve_name()` and `minversion()` in from the main astropy repo? The main astropy repo is a requirement for running anything in this repo, so shouldn't we just import them from astropy?

cc @taldcroft

adrn:
@x-calibre Ah, maybe it's because you want to check if the user has an appropriate version of astropy rather than dying on a blunt `ImportError`? What if we add something to the the top like:

```python
try:
    import astropy
except ImportError:
    raise ImportError("Astropy is a core dependency of this package. See the README file "
                      "or pip-requirements.txt for more information.")
```

x-calibre:
Thanks for reviewing again @adrn !
In one of the previous commits, @taldcroft had suggested that beginners might fork this repo and straight away try to run the tutorials without looking at the requirements.So we thought it was better if we could have the `check` script independent of the environment. I had written the same piece of code(checking for ImportError)  but then later edited it out as Astropy was being checked once for the `requirements.json` file also. 

Should I add in the above code and remove the `resolve_name()` and `minversion()`? @taldcroft @adrn

adrn:
I'm in favor of removing them and just adding a soft failure if Astropy isn't installed -- that is a core dependency for all of this code...

taldcroft:
@adrn - that sounds OK to me.

x-calibre:
Modified as per the suggestions ! Should be ready to merge @taldcroft @adrn  ! It uses `astropy.utils.introspection.minversion` as required and puts a soft check for astropy in the beginning ...

check_env.py
```diff
@@ -0,0 +1,120 @@
...
+                        unicode_literals)
+# Standard library
+import os
+import json
+
+try:
+    import astropy
+except ImportError:
+    import logging as logger
+    logger.error("'Astropy' package is missing and is a core dependency to run this script.\
```
adrn: Both here and below, I would just re-raise an `ImportError` but with your custom message, rather than using `logger.error()`, e.g.,

```python
try:
    import astropy
except ImportError:
    raise ImportError("'Astropy' package is missing and is a core dependency "
                      "to run this script. For more information visit "
                      "http://www.astropy.org/")
```

check_env.py
```diff
@@ -0,0 +1,120 @@
...
+    quit()
+    #Astropy is necessary to run tutorials of astropy. If not present, stop immediately
+
+
+from astropy import log as logger
+
+try:
+    import six
+except ImportError:
+    logger.error("Python library 'six' is missing and is required to run this script.\
```
adrn: Here as well, re-raise an `ImportError`.

x-calibre:
Thank you so much @adrn for reviewing the code again and suggesting changes ! The changes have been incorporated.

PR #84, opened 2015-04-18
Update the deploy script to make it easier with non "origin" remote

eteq:
This updates the ``deploy`` script basically just to make it easier when the remote for astropy isn't called "origin" - it now prompts for a name if "origin" isn't avaialble. (I use "astropy" as the remote so I don't get confused between the "eteq" and "astropy" github accounts...)

cc @adrn as I think the main author of the script


adrn:
Makes sense to me!

PR #85, opened 2015-04-22
Cosmology tutorial showing how to plot something vs both redshift and universe age

nhmc:
Often people want to plot versus redshift and the corresponding age of the universe. This tutorial shows how to do this using `astropy.cosmology` and matplotlib. 

You can view the notebook [here](http://nbviewer.ipython.org/github/astropy/astropy-tutorials/blob/8c0d34157862474d683c3463ccb27ef660297d45/tutorials/cosmology/Cosmology.ipynb).


mwcraig:
@nhmc -- looks good! There were a couple missing imports (numpy, astropy.units), a few typos, and I think matplotlib now handles astropy units. I put a rendered version of comments at http://nbviewer.ipython.org/gist/mwcraig/40f416ad4b53d6afc4ed 

My comments are in the markdown cells in red.

nhmc:
Thanks. I think I fixed everything, here's the [new version](http://nbviewer.ipython.org/github/nhmc/astropy-tutorials/blob/cosmology_tutorial/tutorials/cosmology/Cosmology.ipynb).

kelle:
The meta data of the notebook needs to be updated...in particular, the link_name, author, and date fields. The link name should be the title of the tutorial which will appear in the list of tutorials. I suggest "Making a plot with both redshift and universe age axes".

nhmc:
Thanks, I've updated this.

nhmc:
I fixed a couple more typos - hopefully this should now be ready to merge.

adrn:
@nhmc Sorry for the delay -- I'll have a look at this today or tomorrow!

adrn:
I'm happy with this! Could you add these lines above any matplotlib imports to make the plot styles consistent with other tutorials? 

    # Set up matplotlib to use a nicer set of plot parameters
    %config InlineBackend.rc = {}
    import matplotlib
    matplotlib.rc_file("../../templates/matplotlibrc")

nhmc:
Ok, done.  The new github ipython notebook rendering is pretty awesome :)

adrn:
I agree -- this makes reviewing tutorials so much easier!

adrn:
Hmm, some thoughts just occurred to me. Right now, the bulk of the code in the tutorial is really just getting the plot to look nice. Maybe we should consider adding a `CosmologyAxes` or `RedshiftDistanceAxes` object to `astropy.cosmology` (for matplotlib) that will automatically plot redshift and distance with nice defaults? Seems like there is a lot of boilerplate code for just getting those two x axes to play nice together...Anyways, that is outside of the scope of this PR, just mentioning here so I don't forget.

For this tutorial though, maybe you could add one more demo of a cosmology feature? For example (doesn't have to be this, but one idea), also load the `Planck13` cosmology and plot the two angular diameter distances vs. redshift to show difference between the two cosmologies?

adrn:
Also, this is just a personal preference, but I tend to put semi-colons at the end of the last line of cells that make plots or assign the output to a variable so you don't end up seeing the output like:

    [<matplotlib.axis.XTick at 0x10878fb10>,
    <matplotlib.axis.XTick at 0x108776fd0>,etc...]

nhmc:
I added semi-colons and included your suggestion of overplotting the Planck 2013 cosmology for comparison.

And yes, I agree the boilerplate code needed for two axes is a pain. That's part of the motivation for this tutorial.

adrn:
Awesome -- thanks @nhmc! I'm happy with this unless there are any other comments from @kelle?

kelle:
Nope. Green light to publish from me!  We can always polish and improve it
later but right now, make it live.

On Wed, May 13, 2015 at 8:59 AM, Adrian Price-Whelan <
notifications@github.com> wrote:

> Awesome -- thanks @nhmc <https://github.com/nhmc>! I'm happy with this
> unless there are any other comments from @kelle <https://github.com/kelle>?
>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/astropy/astropy-tutorials/pull/85#issuecomment-101654565>
> .
>



-- 
Kelle Cruz, PhD — http://kellecruz.com/
917.725.1334 — Hunter: x16486 — AMNH: x3404

adrn:
Thanks @nhmc

nhmc:
Ah, I just realised I introduced a mistake by comparing to the two cosmologies. The conversion from redshift to age is different for them both, so we actually need another set of age axes for the Planck 2013 cosmology. This seems a little complicated, so I think it's best just to remove the comparison with Planck for now.

eteq:
@nhmc - hmm, I see your point on this, but I wonder if this is a good learning opportunity?  That is, you can leave what's in there now, but then add another cell that says "oh wait, but that's *wrong*, and here's why:".  For me, at least, "gotchas" like that often stick in my memory better than simple "here's how to do a thing" sort of tutorials.  Alternatively, perhaps some other kind of plot should be added that compares two cosmologies?  I really like the fact that right now the tutorial shows how easy it is to use cosmologies via the same interface.

Also, this is super minor, but, @nhmc, I wonder if the tutorial directory/file should be renamed "redshift_plot" or something?  It seems a bit restricted in content to be called "Cosmology".

Also, @adrn, it looks like this isn't on the web site - did you run the deploy script?  If not, perhaps we should wait until @nhmc finishes those updates?

adrn:
@eteq correct -- I didn't push it to the site because I saw @nhmc's comment -- waiting for revision.

PR #92, opened 2015-05-20
Add travis build to check that notebooks run

adrn:
This adds the necessary travis config files to have travis check that the notebooks actually run properly using `python prepare_deploy.py run`.

@eteq does this look right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: python
+
+os:
+    - linux
+
+env:
+    global:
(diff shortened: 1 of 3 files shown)


eteq:
yep, looks right except that we *might* want to add a second build for py 3.4.  First lets see if this one suceeds, though.

adrn:
Phew, finally got all the right dependencies in to the install script...

adrn:
(PS: I'll merge this tomorrow unless @eteq has any other comments)

PR #96, opened 2015-06-28
fix some simple typos in UVES tutorial

hamogu:
Two things to note:

- I did not review the full notebook. I just looked over it and fixed a couple of simple typos in the text or code.
- I'm running a newer version of IPython, so unfortunately this upgraded to notebook version 4 when I saved it, thus my edits might be hard to find between the lines that are just format changed, but I promise it's only better (word order, some comment lines had wrong indentation in code, one variable was missspelled causing a "show full traceback" button, markdown rendered a "*" as italics instead of "multiply").

Before merging please make sure the IPython that renders the production version (I have no idea how you set that up) works with this version.

### DIFF
--- a/tutorials/UVES/UVES.ipynb
+++ b/tutorials/UVES/UVES.ipynb
@@ -1,1384 +1,1412 @@
 {
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Analyzing UVES Spectroscopy with Astropy"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Acknowledgments"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "License: This tutorial is made available under the terms of the\n",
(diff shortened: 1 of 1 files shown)


hamogu:
@adrn @eteq : Do you have write permission in this repro? If so, can you review and merge my PR or say "We don't do that, because we don't want notebook of version 4 because it's too new?"

adrn:
Sorry, this is going to have to wait a little longer. Things have been slow in the astropy-tutorials the past few months (mostly my fault) but now will grind to a halt going into job season...will get back to this when I resurface. Thanks for the PR though!

eteq:
@hamogu - I went ahead and took care of this.  I think it's best for now to stick with v3 because plenty of people won't be up-to-date enough for v4, IMHO.

Fortunately, there's a straightforward way to backtrack - you just do: ``ipython nbconvert --nbformat=3 --to notebook UVES.ipynb``, which creates a ``UVES.v3.ipynb``.  I then just copied back onto UVES.ipynb, removed a few extraneous pieces of metadata that nbconvert created, and then merged that into master.

So if this happens again, hopefully you can use that trick to move it back to v3.  Sorry it all took so long!

PR #102, opened 2015-12-01
Auto deploy machinery

### LINKED ISSUE #93: Travis-based auto-deploy machinery
I thought there was already an issue for this but I can't fine it (maybe @adrn can confirm?).  It would be great if we can make travis automatically *deploy* the "published" tutorials. 

I think this may be pretty easy to do using https://github.com/ewanmellor/gh-publisher - (which I stumbled upon thanks to http://jakevdp.github.io/multiband_LS/)

eteq:
This should close #93 by finally setting it up so that merging a PR into master auto-deploys to github pages.  

The main idea is that this modifys the ``deploy`` script to allow non-interactive use, as well as the @astropy-bot account as a user that has the appropriate access privileges along with a travis-encrypted access token. 

The awkward thing is that we can't test it really until merging, as it's specifically *not* supposed to work until merged into master.  If it helps any, though, you can see the results of this by looking at https://github.com/eteq/astropy-tutorials/tree/gh-pages, which was made by the build you see at https://travis-ci.org/eteq/astropy-tutorials/builds/94067132 .  This should be basically the same, except with new tokens and such as relevant for the real astropy-tutorials repo.  (Also, a bit more security that should completely hide the secret token).

cc @adrn @astrofrog

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,18 @@ env:
         - NUMPY_VERSION=1.9
         - PYTHON_VERSION=2.7
         - ASTROPY_VERSION=1.0.2
+        - GH_REF: github.com/astropy/astropy-tutorials.git
+        - secure: "NBZlEg//sDYvJikqXESLsx/WA1Nj7fAH2mJfI8ZVMnolvyjJuLdEMxxEYFP8t6U/iieSSrgUkgP94yy5Oq6iJRjj7ThGCdcjgvgWTQCzoQgEVDc5SzdJGyGs30VC+8t7GTobzMe086HX16SJRvsIumqXdmRcJjAdMCyhMUxLoUg="
 
 install:
     - source .travis/setup_environment.sh
 
 script:
     - python prepare_deploy.py run
+
+deploy:
+    provider: script
+    script: ./travis_deploy
+    on:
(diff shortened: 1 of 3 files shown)


adrn:
Nice! Thanks for this!

BTW: one thing that is probably unrelated to this PR that I just noticed -- the redshift plot tutorial .html file is missing a letter `edshift_plot.html'

astrofrog:
Very cool! No objections from me. One idea for testing before master would be to deploy to another branch if we're not on master, and then view it locally. For instance, you could push to a branch called ``<branch_name>-preview`` and then check that out locally if you want to see what it looks like?

eteq:
@astrofrog - ah, clever idea, but it turns out that travis doesn't allow it.  From https://docs.travis-ci.com/user/deployment/#Pull-Requests:

> Note that pull request builds skip deployment step altogether.

What I *can* do is modify this to use a test branch, then we can merge it into master, check if it's working and if so, then modify it only to change to the real name.  Will push up another commit that does that momentarily.

eteq:
Alright, I'll merge this now and we'll see if it creates a `test-gh-page` branch that looks right.  If it does, I can change it to the "real" gh-pages in master.

PR #105, opened 2016-03-24
Changed FITS-table to use a different events file

eblur:
The FITS-table tutorial will now create an image from a list of Chandra events.  It is much more intuitive than the previous example.


eteq:
Looks like this includes commits from when this was first created - I think you only want commits e1006c67981423ef18d12a982812188c99810bda and later, @eblur.

Rebasing notebooks can be really difficult, though.  So it might be easiest to  create a new branch from the current astropy master, and then just copy from this one the file you want changed (I think it's just tutorials/FITS-tables/FITS-tables.ipynb?) and commit that on its own.  Then re place this branch with that. You'll lose the commit history of your changes here, but it looks to me like most of your commits don't necessarily stand alone, anyway?

PR #108, opened 2016-04-20
Updated FITS tutorials

eblur:
Yup! They are updated. Hopefully.


eteq:
Note that this should be the same as #105, but with the revision history cleaned up.

eteq:
I just pushed up the "built" version of these to http://eteq.github.io/astropy-tutorials/ (with a small change to make it work even though we haven't yet merged astropy/astropy-data#12), which you can compare to http://tutorials.astropy.org .

@eblur - a few comments:

1. I don't see any changes to "Viewing and manipulating FITS images", but the git changes here show a change to that file... Is that expected?
2. The cell in the new version labeled ``Out[12]`` seems to be very different from ``Out[14]``.  Any idea why?
3. Would you consider changing the color map for ``In[14]`` from "jet" to "viridis"? See https://twitter.com/astrofrog/status/623088520740118533 for reference ;)
4. Comparing the the [new](http://eteq.github.io/astropy-tutorials/FITS-tables.html) and [old](http://www.astropy.org/astropy-tutorials/FITS-tables.html) versions of the table tutorials, I see the new one has one fewer exercise. Of course that's because the new version sort of does what the old one's question is about... but perhaps you can come up with a similar question that's a bit more in the "explore the data" rather than "re-visualize the same data in a new way"?

pllim:
Re: viridis

If you plan to cater to people having Matplotlib 1.3, Viridis is not a good choice because it doesn't exist back then. FYI.

eteq:
Oh, and I think I may have figured out 1 on my own: it looks like the notebooks you checked in here have been *executed*.  The actual tutorial notebooks in the repo should be un-executed (i.e. what you get if you do "cell->All Outputs->Clear" in the notebook).  

The notebook outputs get generated as part of the deployment process, which helps make sure they actually run correctly (And it also means we don't have to save the images as part of the notebooks).

eblur:
1. When I was running the FITS-image tutorial on my laptop, I got an error when trying to make the histograms. So I changed .flat to .flatten()
2. I'm not sure...
3. Yup, changed it.
4. Oops, saw that just now. Added a question.

eblur:
Are there any other changes needed?  I just noticed that this hasn't been closed yet.

eteq:
@eblur - oh, I thought you were still working on it so hadn't looked back at this.  Will take another pass as soon as I get a chance (and hopefully be able to merge it!)

eblur:
bonk

eteq:
@eblur - eblur/astropy-tutorials#1 has some minor updates, after which I think this is *almost* there.

However, when I try running the table tutorial, I see a problem: it appears the data file is *different* from what the tutorial was design for.  Specifically, the filtering on ``'ccd_id'`` yields *no* events in this version of the tutorial (the file only has ccd_id 5,6,7,8).  Any idea what changed?

eteq:
(note that a quick fix would be do just change the ``0, 1, 2, 3`` to ``5, 6, 7, 8``... But that makes the plots look different.  Not sure how important that is, though.)

eblur:
Oh, it's actually important.  The old file was ACIS-S, which have ccd_id numbers of 4,5,6,7,8,9

The new file is ACIS-I, which has ccd_id numbers of 0,1,2,3

We want the new tutorial to use the new data file.  I thought we updated astropy tutorial files for this. Let's figure it out by the end of the week.

eteq:
Aha!  It's because astropy/astropy-data#12 needs to be merged!  I just tested this with this, and I think it's now working.  Will merge them both now.

PR #112, opened 2016-11-08
update coordinates tutorial to have a py 2/3 compliant use of urllib

eteq:
Just what the title says... Previously it only worked in py 2 (as pointed out by @astrofrog)

The diff is weirdly large, though ... the only content change is the change to the cell that import urllib (and where it's called)... I'm not sure why so much of the file changed.  Maybe a new nb format?

cc @astrofrog @adrn


eteq:
I'm just going to merge this - hopefully it will be fine.

PR #124, opened 2017-05-12
added more content instructions and polished

kelle:
Based on discussions held at #pyastro17 unconference session.


adrn:
This is docs-only, so canceling travis and merging

PR #135, opened 2017-08-04
some minor grammatical changes

MycrofD:
This is a minor edit for grammatical changes. Shouldn't be much of a problem


adrn:
Thanks for providing these changes - we're now hitting a ton of conflicts, and the effort to resolve might outweigh the effort required to just re-make the changes! But: we're also going through and auditing all of the content, so we'll try to address these specific grammar issues later.

PR #153, opened 2017-09-22
Switch tutorial infrastructure to use Sphinx

adrn:
This is a major overhaul of the build/deploy machinery underlying the astropy tutorials site. The core idea is to move from our (custom, broken) scripts for running and deploying the tutorial content as static pages. This instead uses `nbsphinx` and the astropy sphinx template to build the tutorials into a website that looks more like the astropy documentation.

[There is a rendered demo of this here.](http://astropy-tutorials.readthedocs.io/en/latest/) 

cc @eteq @astrofrog @bsipocz @kelle


bsipocz:
why do we need astropy_helpers here, is it only for the css?

adrn:
Yea, AFAIK

adrn:
OK, you should be able to check out my branch and run `python setup.py build_docs` and see the new site. See discussion items / questions listed above though

bsipocz:
Would be ``python setup.py build_docs`` enough as a test in our CI? I guess we both want to test the content and whether it builds, don't we? Is ``setup.py test`` still works or it's now something else?

(I'm asking so that it's not dummy script that I put in the travis framework POC PR)

bsipocz:
In case of narrative docs it doesn't checks the content. Does it work that differently for notebooks?

adrn:
I just found out about [nbval](https://github.com/computationalmodelling/nbval) - that might be the solution, and just use nbsphinx to build/display the tutorials?

kelle:
Any chance we could have a brief 30 min hangout to discuss this tomorrow? I'll be in Madison with @eblur prepping for the two day sprint which starts on Thur.

adrn:
Even better: I'll be at the CCA. But the short answer is that we won't have this ready in time. But with this, we can still accept tutorials as notebooks, so for new content just follow the same protocol.

adrn:
Oops I misunderstood - you're already in Madison. Ok yes hangout tomorrow would be fine

adrn:
As @eteq pointed out in a chat window, it looks like nbval operates by comparing pre-cached output to output generated from re-running the notebook. Currently, the notebooks stored in the repo have their output cleared, so we would have to change our routine if we went with this route...

But @bsipocz: I didn't follow your point
> In case of narrative docs it doesn't checks the content.

adrn:
Ugh, just noticed something else undesirable...

In many of our notebooks, we have links that are formatted like code, e.g., [`Quantity`](...). These don't convert to rst because, apparently, [rst doesn't support nested inline markup!!!!](https://stackoverflow.com/questions/4743845/format-text-in-a-link-in-restructuredtext). To quote a comment from that link:

>This is horrible. All sane markup languages support indefinite nesting of inline markup, because this is 2016. Context-free parsing has been trivially solved since 1959. (Chomsky: "On certain formal properties of grammars.") The inability of the reStructuredText parser to perform genuine context-free parsing is a crass, ugly blemish on an otherwise sterling facade. The well-defined and highly extensible syntax of reST deserves better.

Of course, #160 would solve this...but that requires some serious engineering...

bsipocz:
>But @bsipocz: I didn't follow your point

>>    In case of narrative docs it doesn't checks the content.

I'm fairly certain (but probably need to double check), that you can build a docs with sphinx that contains un-runnable code snippets. Now, this may be not true for nbsphinx.

adrn:
With a hack to rule all hacks, I was able to get a link to the source notebook to appear on the bottom of the rendered page! Here's why it was so hard (documenting for myself, please don't read this, you really don't want to know)

(seriously, you don't want to read this far)

`nbsphinx` uses `nbconvert` by defining a custom RST exporter for ipython notebooks. So, during the sphinx build process, sphinx calls the `nbsphinx` build process, which executes the notebooks and calls the `nbsphinx` custom RST exporter to turn the executed notebooks into RST files. The RST files then get grabbed by sphinx and built into the docs as normal. `nbsphinx` defines its own custom jinja2 template for `nbconvert` to customize how the notebooks are turned into RST. But, `nbsphinx` (the sphinx extension / package) is just a single python module. i.e. it's not in package layout, meaning they define everything in that single [nbsphinx.py](https://github.com/spatialaudio/nbsphinx/blob/master/nbsphinx.py) file. That includes the custom jinja2 RST template. If they used a package layout, and made that a file, we could subtemplate the nbsphinx template and customize whatever we wanted (like, e.g., adding a link in the footer section). But that's not possible with the current layout. So instead we [monkey patch](https://stackoverflow.com/questions/5626193/what-is-monkey-patching) the `nbsphinx.RST_TEMPLATE` variable in our `conf.py` file to contain our custom template with the link to the source notebook. See the bottom of [astropy.tpl](https://github.com/adrn/astropy-tutorials/blob/sphinx/docs/_static/astropy.tpl) to see the hacks in the footer section.

cdeil:
@adrn - What's the status / plan here? I think the build / deploy for the tutorials in master is broken anyways, so maybe it could make sense to already merge this branch now, if it's the way to go and it's mostly working?

The reason I'm bringing this up is that @bultako volunteered to implement an nbsphinx based solution for the Gammapy tutorial notebooks:
https://github.com/gammapy/gammapy/pull/1168#issuecomment-337589087
https://nbviewer.jupyter.org/github/gammapy/gammapy-extra/blob/master/index.ipynb

And naturally, as an Astropy-affiliated package, we're looking at what Astropy does for guidance and would like to adopt a similar tooling / setup. :-)

My main questions are:
- do you want to go this way, use nbsphinx?
- is this a separate Sphinx build, or integrated with the Sphinx build of Astropy somehow?
- do you have a good solution / plan already how to run and deploy?

adrn:
@eteq and I are going to sprint on this tomorrow afternoon - hopefully we'll tie up loose ends at that point. 

> do you want to go this way, use nbsphinx?

It did seem like the way to go, but now I'm thinking of doing something a little more customizable. More on this after I try out a few ideas...

> is this a separate Sphinx build, or integrated with the Sphinx build of Astropy somehow?

This is a separate sphinx build.

> do you have a good solution / plan already how to run and deploy?

I had an idea (based on @dfm's own solution - see below) but need to try it out before I'm sure it's feasible. More on that soon...

BTW: @dfm just pointed me to his own build setup that doesn't use nbsphinx but does similar things:
> Here's my hacky setup for building tutorials:
> I save the tutorials in docs/_static/notebooks so that they will be downloadable. Then I added this rule to my makefile:
> https://github.com/dfm/emcee/blob/master/docs/Makefile#L23
> This runs nbconvert on all the tutorials and saves them to docs/tutorials.
> It also runs this script to fix the internal links:
> https://github.com/dfm/emcee/blob/master/docs/fix_internal_links.py

cdeil:
> This is a separate sphinx build.

Do you have a solution to link from the markdown cells in the notebooks back to Astropy Sphinx docs? Are you using `:ref:` and intersphinx for this, or just hard-code the link to the full URL, or avoid that completely?

What we have in Gammapy at the moment is that we try to link to the relevant API docs of the classes we're using, like to `SkyImage` [here](https://nbviewer.jupyter.org/github/gammapy/gammapy-extra/blob/master/notebooks/first_steps.ipynb#Sky-images). At the moment we do this just by linking to the full URL, which is not a good solution (annoying to write). But it's hard to come up with a better one, if we want to keep using Readthedocs and store pre-rendered notebook in the repo that are just converted, not executed by nbsphinx (otherwise it would be too slow, executing all notebooks on RTD would time out).

> I had an idea 

So we're very interested in what you use in the end. For Gammapy, we talked a bit today and plan to use nbsphinx, running on RTD, converting pre-executed notebooks that are stored with output in the repo where we maintain the notebook, because that seems like the simplest solution to get something running now, and to get benefits like navigation & search for the notebooks. And maybe, hopefully, easier x-linking between notebooks and Sphinx docs too.

dfm:
@cdeil: with my method, I just write the usual `:ref:`, `:class:`, etc. in the notebooks and then I automatically run [this script](https://github.com/dfm/emcee/blob/master/docs/fix_internal_links.py) (after nbconvert) to reformat them work for internal references.

cdeil:
> with my method, I just write the usual :ref:, :class:, etc. in the notebook

That means that while editing the notebook, the links don't work for you, and if a user downloads your notebook, the links also don't work, right?

For Gammapy the notebooks aren't just used to write docs, a major use case is that users can download and read and execute them. Like https://github.com/jakevdp/PythonDataScienceHandbook , but in addition we want `:ref:`, and also generate IPYNB files that have working links, for users to execute and read. I don't think it's easily possible, but we'll try.

@dfm - Where are your converted notebooks? I can't find them on your docs page!?

Side comment:  the link to " quickstart documentation" at http://emcee.readthedocs.io/en/stable/#basic-usage is currently broken.

dfm:
Yeah you're right. Mine doesn't inject links into the notebooks. I can see ways to do that, but it's pretty ugly!

I'm in the middle of a docs update so you can find the notebooks here: http://emcee.readthedocs.io/en/latest/

cdeil:
Ah, I see. So the links are resolved when a user views your docs online, but if I download and start reading / executing the notebook I see
```
:func:`EnsembleSampler.get_chain`
```
unresolved / without working link to your API docs.

I think for now, we'll end up with the same situation for Gammapy.

And then in the future, we could add the step where we also generate new IPYNB files, with those links resolved. It's probably hard to figure out how to load the target resolution info from Sphinx, and how to modify the IPYNB via the programmatic interface in nbconvert, but in the end it should be possible with a few lines of code, no?

It looks like http://nbsphinx.readthedocs.io/ doesn't aim to emit IPYNB at all, so probably that part would have to be scripted separately.

---

It's still not clear to me why you don't go with `nbsphinx`?

Can you point out a specific issue with trying to use it, where it's not possible or too hard to customize something? Or is it just a matter of "don't need it, prefer to roll my own"?

Bultako:
Hi
I think we could easily replace absolute links with relative links tuning the script [fix_internal_links.py](https://github.com/dfm/emcee/blob/master/docs/fix_internal_links.py) with new regular expressions, or maybe I am missing something.. At least, that was my initial idea..

I'm easily converting a mixed doc of *.ipynbs and *.rst file into gammany sphinx formatted html docs with `nbsphinx`, in just one shot of `sphinx-build`

.... but I am blocked with warnings in the `:code`directive, which fails in the insertion of code :( If I get to solve this I think we will have a nice seamless process to integrate notebooks and RST in the same doc.

adrn:
Thanks all for the ideas. I added an updated plan to the PR description above. If you have any thoughts, add comments here or use the #learn slack channel in the astropy slack.

adrn:
There's now a `convert.py` script in `scripts`. You can run all of the notebooks with:

```bash
cd scripts
python convert.py -v ../tutorials/ --exec-only
```

or execute individual notebooks with, e.g.,
```bash
python convert.py -v ../tutorials/coordinates/coordinates.ipynb --exec-only
```

I've added the cell metadata tag to each of the cells we expect to fail (i.e. the ones that intentionally raise exceptions), so the execution should succeed. If you also want to try generating the RST files, just remove the `--exec-only` flag:

```bash
python convert.py -v ../tutorials/
```

This is almost complete (the hard part is done). The things that are left to do:
[ ] Add the notebook names to the notebook metadata (we'll need this so we can auto-generate links to the source notebooks).
[ ] Use a custom template in the nb -> RST converter so we can add things to the generated pages. Specifically, I think we want to link to (1) the source notebook to download, and (2) a live version of the source notebook on Binder.
[ ] Configure travis and decide on the builds to do on travis. I think we want to (1) just execute all of the notebooks to see that they run, and (2) execute the notebooks, convert to RST, and try the sphinx build
[ ] Configure readthedocs.

cdeil:
One potential issue with this setup that could cause problems in the coming years is that there is no explicit version coupling between the two repos: `astropy` with the code and Sphinx docs, and `astropy-tutorials` with the notebooks and scripts for the notebook build. There's also `astropy-helpers` with Sphinx and tooling code, but that is version-coupled via a git submodule. And I presume for you the data files the tutorials load are fixed and you can assume they will be there and unchanged in a few years (something that's not the case for us in Gammapy where we are simultaneously working on new file formats).

So concretely: what if someone has to reproduce the Sphinx and tutorial docs for some old branch (say 2.1 or whatever, something a few years old), e.g. because the RTD servers were hit by a meteor?

It might be helpful to couple the versions of the repos explicitly, e.g. via another git submodule pointer either from `astropy` to `astropy-tutorials` or the other way around?

cdeil:
PS: I just now merged https://github.com/gammapy/gammapy/pull/1176 adding notebooks via `nbsphinx` to the Gammapy Sphinx docs, so this is the way we're heading ...

adrn:
The tutorials will ship with pinned requirements (via a conda environment file and a pip requirements file) and will be released (tagged) in parallel to astropy. If someone wants to use an old version of astropy and an old version of the tutorials, the versions should sync. That's as much as we're going to promise, but I don't know what the alternative is in this paradigm - are you  arguing that the tutorials should live in the astropy core repo? That can't happen because they will grow to include tutorials that span the astropy ecosystem rather than just the core package.

adrn:
Successful build on readthedocs! I think this is ready for review @eteq @kelle 
http://astropy-tutorials.readthedocs.io/

cdeil:
> are you arguing that the tutorials should live in the astropy core repo? 

No, I was wondering if it makes sense to version-couple the `astropy` and `astropy-tutorials` repos via a git submodule pointer for you.

> The tutorials will ship with pinned requirements (via a conda environment file and a pip requirements file) and will be released (tagged) in parallel to astropy.

Ah, that wasn't clear to me, that you want to couple the versions this way.

It looks like you currently have almost no links back and forth between the astropy Sphinx docs and the `astropy-tutorials`, and the few that you have on the front page are just to generic URLs that will likely never change and aren't versioned.

So yes, if you don't have links back and forth between the two Sphinx docs, then what you have here will work fine I think.

In Gammapy we have many links back and forth between the notebooks and the Sphinx docs (see e.g. [here](http://docs.gammapy.org/en/latest/notebooks/first_steps.html)). In that case it becomes difficult to have separate Sphinx projects that link back and forth (e.g. via `intersphinx`). I was thinking we attempt to solve that issue by adding the repo with the notebooks as a git submodule in the main repo and have one Sphinx build there (not sure if this will work nicely yet).

bsipocz:
> I was thinking we attempt to solve that issue by adding the repo with the notebooks as a git submodule in the main repo and have one Sphinx build there (not sure if this will work nicely yet).

The docs build time is already so long for astropy, that at the brainstorming session we definitely looked for ways to do the testing/building of tutorials separately (in addition to the other issue @adrn mentioned, that the tutorials is more about the Astropy Project than the core package). E.g. we also came to the conclusion to remove the sphinx-gallery examples, too. 
At some point I hope that the upstream packages add the possibility of using astropy's intersphinx (though doing it both ways may be problematic).

eteq:
Re: https://github.com/astropy/astropy-tutorials/pull/153#issuecomment-338496964

> So yes, if you don't have links back and forth between the two Sphinx docs, then what you have here will work fine I think.
> In Gammapy we have many links back and forth between the notebooks and the Sphinx docs ...

I think we may *eventually* want some links back to the tutorials... But @cdeil, can you clarify why you can't depend on linking to the URLs of the tutorials rather than depending on `intersphinx`?  E.g., you can link to `https://nbviewer.jupyter.org/github/gammapy/gammapy-extra/blob/master/notebooks/first_steps.ipynb` instead of using the sphinx machinery?  That's how I had been assuming we'd resolve this if we want to start linking into the tutorials from the docs...

(On the actual PR, looking very nice!  Have some stylistic thoughts on how things are rendered, but will probably just issue a PR against @adrn's branch to address those.)

cdeil:
@eteq - As soon as you have to things that depend on each other, and you don't have a system to link their versions, you have a problem, no?

So e.g. from the tutorials, would you start linking to http://docs.astropy.org/en/v2.0.2/ or http://docs.astropy.org/en/latest/ or ... ?

I think for Gammapy the solution to make `gammapy-extra` which has the notebooks a git sub-module of the `gammapy` code repo and to have a single Sphinx project that includes and offers the notebooks for download solves the issue (at the cost of frequently having to update the git submodule pointer).

adrn:
> It looks like you currently have almost no links back and forth between the astropy Sphinx docs and the astropy-tutorials

@cdeil I'm planning to use the script @dfm shared above to fix `:ref:`-style links so that intersphinx will work as expected. We'll just have to go through and add more links. We could also consider doing what sphinx-gallery does, which is to replace any `package.module.function` calls with a link via intersphinx back to the source docs. But that'd be a follow-up PR.

> As soon as you have to things that depend on each other, and you don't have a system to link their versions, you have a problem, no?

Maybe I don't understand, but intersphinx [lets you control the version](http://www.sphinx-doc.org/en/stable/ext/intersphinx.html) that the links point to. So as long as we increment that with the version of the tutorials, and as long as that is in sync with the Astropy core release version, I think we'll be all set without having to use submodules.

cdeil:
> So as long as we increment that with the version of the tutorials, and as long as that is in sync with the Astropy core release version, I think we'll be all set without having to use submodules.

Yes, I think you're right.

But that's painful to maintain: having to manually update version numbers and tags in multiple places in the right order. But updating git submodule pointers is also annoying. Since you have two separate Sphinx builds I'm not suggesting any more you implement that.

👍

Bultako:
Hi there
We have an [issue](https://github.com/gammapy/gammapy/issues/1184) in `gammapy` RTD documentation related with the search functionality.  

It seems RTD performs the search with a server-side API that [searches in json files](http://docs.readthedocs.io/en/latest/yaml-config.html?highlight=JSON) instead of the local JS search provided by Sphinx. This means that searches do not provide the same results when performed in local Sphinx doc than in RTD. In `gammapy` the difference is specially relevant when trying to search for notebooks. The local search provided by Sphinx works nicely and provides accurate results. I don't know if `astropy` RTD docs will have the same problem as `gammapy`

Some RTD users have proposed to fix this [injecting a JS file in the layout.html](https://github.com/rtfd/readthedocs.org/issues/1487#issuecomment-323290547) theme file, which in principle would perform the search in the user's browser using the downloaded JSON blob produced by Sphinx.

Since `gammapy`is using the `astropy.helpers` as a git submodule to build the docs, do you think this proposed fix could be explored?

- [extra.js](https://github.com/nodemcu/nodemcu-firmware/blob/7dd89dd15ef993c9740ba4d4361e6be8edd1284b/docs/js/extra.js) file to add in _static folder

- {% set script_files = script_files + ['_static/sidebar.js'] + ['_static/extra.js'] %} in layout.html

adrn:
Hey @Bultako. Sorry, I don't follow. To me it looks like the search works as I would expect on the astropy-tutorials RTD page - do you see otherwise, or is the issue you raise only pertaining to gammapy?

Bultako:
@adrn 
Yes you're right, I thought `astropy-tutorials` were still not deployed in RTD, I just checked the link Tutorials in http://docs.astropy.org, sorry my fault. 

As you say the search works pretty well in the astropy notebooks. Did you have any time out issue when building `astropy-tutorials` in RTD? We had in gammapy and I suspect the JSON files which RTD uses for the search were not correctly built.

adrn:
So far, no, the tutorial build process seems to be dominated (in time) by the conda installation.

adrn:
This PR at least works (so, better than what's in master by my eye). 

If @eteq hasn't reviewed by Thursday, I'll merge and ask for issues or a follow-up PR to edit! 

The agreement:
image

scripts/convert.py
```diff
@@ -0,0 +1,244 @@
...
+logger.setLevel('INFO')
+
+from nbconvert.preprocessors import ExecutePreprocessor, CellExecutionError
+from nbconvert.exporters import RSTExporter
+from nbconvert.writers import FilesWriter
+import nbformat
+
+IPYTHON_VERSION = 4
+
+class NBConverter(object):
```
eteq: Maybe call this ``NBSphinxTutorialsConverter`` or something?  Right now it looks like it's part of `nbconvert` itself or something

scripts/convert.py
```diff
@@ -0,0 +1,244 @@
...
+                                               'don\'t convert them as well. '
+                                               'This is useful for testing that'
+                                               ' the notebooks run.')
+
+    parser.add_argument('-o', '--overwrite', action='store_true',
+                        dest='overwrite', default=False,
+                        help='Re-run and overwrite any existing executed '
+                             'notebook or RST files.')
+
+    parser.add_argument('nbfile_or_path', default=None,
```
eteq: Maybe this should default to ``docs/_static/tutorials``?  (Or ``.``?)

scripts/convert.py
```diff
@@ -0,0 +1,244 @@
...
+        # Execute the notebook
+        logger.debug('Executing notebook...')
+        executor = ExecutePreprocessor(timeout=900, kernel_name='python3')
+        with open(self.nb_path) as f:
+            nb = nbformat.read(f, as_version=IPYTHON_VERSION)
+
+        try:
+            executor.preprocess(nb, {'metadata': {'path': self.path_only}})
+        except CellExecutionError:
+            # TODO: should we fail fast and raies, or record all errors?
```
eteq: I think "record all errors" is better, but that can await a later PR if it seems too much work to manage right now.

adrn:
On the comments you list above:

> With the removal of the DEPLOY.md there's now no instructions for a local build visible without actually going to the page. Something about this should go in the README, I think.

How about a link to the docs page from the README?

> scripts/check_env.py doesn't work and needs fixing

Yea. I think that could be handled in a separate PR...

> The actual nbs are now in a weird place. Maybe re-do the "docs" heirarchy to have it be the base of the repo?

I would prefer just adding a symlink so that a top level "tutorials" links down to where they actually are.

.travis.yml
```diff
@@ -1,29 +1,52 @@
...
+
         - PYTHON_VERSION=3.6
-        - SETUP_XVFB=True
-        - CONDA_ENVIRONMENT=conda-environment.yml
-        - CONDA_DEPENDENCIES='jupyter notebook ipython runipy'
-        - GH_REF: github.com/astropy/astropy-tutorials.git
-        - secure: "NBZlEg//sDYvJikqXESLsx/WA1Nj7fAH2mJfI8ZVMnolvyjJuLdEMxxEYFP8t6U/iieSSrgUkgP94yy5Oq6iJRjj7ThGCdcjgvgWTQCzoQgEVDc5SzdJGyGs30VC+8t7GTobzMe086HX16SJRvsIumqXdmRcJjAdMCyhMUxLoUg="
+        - NUMPY_VERSION=stable
+        - ASTROPY_VERSION=stable
+        - EVENT_TYPE='pull_request push'
```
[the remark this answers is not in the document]
bsipocz: @eteq - I would suggest to skip reviewing the travis setup as I plan to open a follow-up PR to set up the stages anyway once this is merged.

eteq:
Note in adrn/astropy-tutorials#7 I addressed most of the non-inline comments except the link color (which I think is non-trivial) and the "actual nbs are now in a weird place".

> I would prefer just adding a symlink so that a top level "tutorials" links down to where they actually are.

Can you clarify *why*?  The current ``docs`` directory seems pointless to me because thats the *only* thing here.  And it's not really the "docs", just the tutorials.

adrn:
> Can you clarify why? The current docs directory seems pointless to me because thats the only thing here. And it's not really the "docs", just the tutorials.

Just for organizational purposes. I don't like having the sphinx config and files with all of the other infrastructure files at the top level. I'm cool with renaming the folder "docs" though...

bsipocz:
what about calling the directory "tutorials"?

README.rst
```diff
@@ -0,0 +1,25 @@
...
+To run the tutorials *locally*, you need jupyter notebook installed::
+
+    jupyter notebook
+
+The tutorials are initially empty of any output.  You can run them by pressing
+"Run All" under the "Cell" menu in the notebook file.
+
+## Dependencies
+
+See the conda environment file or pip requirements file for a list of
```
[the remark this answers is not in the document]
adrn: @astrofrog made both - maybe it's just to give people the option of package manager? But agree it's slightly more overhead for us. I'll leave for now.

scripts/convert.py
adrn: 👍

adrn:
Almost there. @eteq 's changes are merged and comments addressed, but still a few things to resolve http://astropy-tutorials.readthedocs.io/en/latest/:
* [ ] The text width needs to be adjusted (too wide right now) - can we use the `custom.css` file?
* [ ] The notebook download links are broken because for some reason they aren't getting copied into `_static`
* [ ] We can remove (or decide whether to remove) the setup.cfg and setup.py files

eteq:
OK, I've pushed some changes that remove the `setup.*` and replace with a Makefile in the *root* of the repository.  So now the process is just cloning the repo, doing `make html`, reading the error about `git submodule`, doing it, and then `make html` again which will *actually* work :wink:

One remaining task: 
- [ ] update the dev docs (and readme) to reflect this new layout

will try to do that ASAP

PR #165, opened 2017-09-29
Coordinate Series [WIP]

stephenpardy:
Starting a pull request of the first draft of a new series on the astropy coordinates library based on the Coordinates tutorial made by @eteq . First step is to split off the different pieces of the original tutorial into their own tutorials. We have started two new ones for now:

Coordinates-Info - creating SkyCoord objects and basic operations. No feedback requested at this time.

Coordinates-Transform - contains the full tutorial on transforming between different coordinate systems, and an basic tutorial of how to transform to observing alt-az.

This is a work-in-progress and comments/suggestions are welcome.


kelle:
Here's the link to the docs about velocities: http://docs.astropy.org/en/stable/coordinates/velocities.html

I think we should add some of this to one of the tutorials unless @adrn or @eteq think we shouldn't.

adrn:
Yes it would be great to have a demo of, e.g., proper motion transformation, or a transformation to the LSR!

kelle:
To Do for Getting Started
- [ ] use astroquery to get the cutout image
- [ ] add another celestial object to show off SkyCoord attributes, such as distance and velocity.
- [ ] transform from distance and radec to galactocentric position

eblur:
- use astroquery instead of url fetching
- i.e. need to learn how to download images with astroquery

eblur:
Resubmitted in PR #179

PR #184, opened 2018-01-23
Update FITS-images.ipynb

sndik16:
Updated the link to the HorseHead FITS image to
'https://astropy.stsci.edu/data/tutorials/FITS-images/HorseHead.fits'


eblur:
data.astropy.org is the official data location that you should use to store things. We cannot assume that it will always be at stsci website. The data.astropy.org link is maintained, the stsci one is not.

PR #198, opened 2018-02-27
Better header

### LINKED ISSUE #197: Move the content of the "green box" tutorial header Note to menu bar
We're planning to have a standard, stylized header / introduction at the start of each notebook. @eblur and I discussed that the current green "note" about Binder could move to make more screen real estate for the new headers. In principle, we could just add two links to the menu bar: "Download" and "Interactive notebook" rather than the current `.. Note::`.

adrn:
Fixes #197 

This identifies any markdown cell that starts with "# Introduction" and re-styles it to have a blue background. The particular string "# Introduction" could be anything, or could even be in the cell metadata.

Also, the style I chose was just to show off that we can control it with CSS, so feel free to suggest other ideas! 

image

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -102,7 +102,7 @@ Template intro
     ## Authors
     Jane Smith, Jose Jones
 
-    ## Leaning Goals
+    ## Learning Goals
     - Query..
     - Calculate..
     - Display..
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -28,11 +28,12 @@ Overview
 * Each tutorial should have 3–5 explicit `Learning Goals
   <http://tll.mit.edu/help/intended-learning-outcomes>`_, demonstrate ~2–3
   pieces of functionality relevant to astronomy, and 2–3 demonstrations of
-  generic but commonly used functionality (e.g., numpy, matplotlib)
+  generic but commonly used functionality (e.g., ``numpy``, ``matplotlib``)
 * Roughly follow this progression:
-    * *Intput/Output*: read in some data (use `astroquery
-      <https://astroquery.readthedocs.io/en/latest/>`_ where possible
-    * *Analysis*: do something insightful/useful with the data
+    * *Input/Output*: read in some data (use `astroquery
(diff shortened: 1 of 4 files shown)


eblur:
Most of our standardized headers start with the title and then give supporting info (e.g. Authors).  So the identifier could be any cell that starts with "## Author"

The question that remains:
Should the title appear before or after the blue box?

@kelle @eteq please comment on this implementation

eblur:
one more comment @adrn 

The "Download tutorial notebook" and "Interactive tutorial notebook" links seem easy to miss. Can we give them buttons or insert them somewhere else that makes them more obvious?

kelle:
If possible,
Title first.
followed by Authors, then Introduction.

I agree that the links/buttons to download and run the tutorial should be more visible...possibly right under the title.
I also think those links/button should be repeated at the bottom of the notebook.

adrn:
After thinking about this a bit more, now I like:
* __Big, bold__: Title
* __Formatted as a normal markdown cell__: authors, keywords , learning goals
* __Formatted differently__ (e.g. with a blue background): Standard setup instructions / intro that should appear on every tutorial (short instructions on how to download, setup, and run locally)
 Title / author / 
* __Normal cell formatting__: tutorial content

So, this would just amount to injecting the standard setup instructions into the rendered pages where relevant.

eblur:
@adrn I like the way you are thinking. However, most of our content doesn't follow that format. It requires putting each piece in a separate cell, correct?

How about for now:
1. Put the Title before the blue box
2. Put the "authors" and all supporting content in the blue box
3. Modify our contributing.rst to ensure that authors put the title and supporting info into separate cells

For later:
We should make a template tutorial file with the header format you suggested above. Once we provide the template, we will (hopefully) populate the website with new and better content. With a critical mass of uniform template styles, we can modify the website style, and also bring the old tutorials up to date with the new header format.

Sound good?

kelle:
that plan sounds good to me.

adrn:
FYI: I descoped this PR drastically to just include better buttons for downloading / going to the interactive tutorial on binder. Will address other matters in other PRs...

PR #201, opened 2018-03-08
Add In/Out markers to generated cells

### LINKED ISSUE #188: Make inputs and outputs more distinct/apparent
In the new tutorial infrastructure, inputs and outputs are difficult/inconsistent to distinguish.  This snippet from the quantity tutorial illustrates the problem:

image

The first code snippet is an input and the second is an output, but you can't tell that without already understanding what's going on.  Further down it's even *more* confusing because a `Quantity` gets rendered without a box, despite being an output of the code snippet above.

Of course it's all clear if you actually go look at the notebook... But I wonder if we want to fiddle around with the default stylings so that these things are more distinguishable? 

cc @adrn @eblur

eteq:
This closes #188 by making a surprisingly simple change to the templates.

This isn't *quite* what I think we wanted, because the In/Out text is *above* the cell instead of to the left of it like a notebook.  But I think it's close enough because it makes the distinction much clearer.

Here's what things end up looking like after this PR:
image

### DIFF
--- a/tutorials/astropy.tpl
+++ b/tutorials/astropy.tpl
@@ -21,9 +21,16 @@
 {% endblock %}
 
 {% block in_prompt %}
+
+*In[{{  cell['execution_count'] }}]:*
+
+
 {% endblock in_prompt %}
 
 {% block output_prompt %}
+
+*Out[{{  cell['execution_count'] }}]:*
+
 {% endblock output_prompt %}
 
 {% block input %}
--- a/tutorials/astropy.tpl
+++ b/tutorials/astropy.tpl
@@ -32,7 +32,7 @@
 
 {% block output_prompt %}
 
-:outputnumrole:`Out[{{  cell['execution_count'] }}]:``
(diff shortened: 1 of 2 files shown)


tutorials/astropy.tpl
```diff
@@ -17,13 +17,23 @@
...
 {% block in_prompt %}
+
+:inputnumrole:`In[{{  cell['execution_count'] }}]:`
+
+
 {% endblock in_prompt %}
 
 {% block output_prompt %}
+
+:outputnumrole:`Out[{{  cell['execution_count'] }}]:``
```
adrn: Extra backtick at end of the line here.

adrn:
Nice! That wasn't so bad. LGTM, except for the small typo I noted in a comment.

tutorials/astropy.tpl
eteq: Done

adrn:
LGTM!

PR #215, opened 2018-04-09
Update tutorials to use new image files

adrn:
*This should only get merged once astropy/astropy-data#48 is merged!*

This updates the new tutorials to download data files from the data.astropy.org repo. Those files are significantly smaller than the originals: I cut out sections of the images to make the files smaller and make the downloads more manageable. They still require downloading >100MB of data...


adrn:
Woot!

PR #225, opened 2018-05-04
adds auto-redirect to old tutorial pages

kelle:
added pink box to top of webpages and auto redirect meta data. Also added versions of old files for archival purposes. Be sure to check auto redirect isn't in the `_old` files.


UVES.html
```diff
@@ -11765,19 +11765,26 @@
...
+
+
 	<!-- Author and Date Information & Edit on Github Link-->
-	
+
+<meta http-equiv="refresh" content="5; url=http://astropy-tutorials.readthedocs.io/en/latest/rst-tutorials/UVES.html" />
+
+<p style='background:pink'>This tutorial have been moved to <a href="http://astropy-tutorials.readthedocs.io/en/latest/rst-tutorials/UVES.html">http://astropy-tutorials.readthedocs.io/en/latest/rst-tutorials/UVES.html</a>.This page should auto redirect you to that page in 5 seconds.<br />
+
+If you really wanted to see the old (unmaintained) version of the tutorials that was previously at this location, it is archived here: <a href =“http://www.astropy.org/astropy-tutorials/UVES_old.html”>http://www.astropy.org/astropy-tutorials/UVES_old.html</a>
```
eteq: I think there's some sort of problem here - when I try to follow this link from the page it sends me to ``file:///Users/erik/src/astropy-tutorials/%E2%80%9Chttp://www.astropy.org/astropy-tutorials/UVES_old.html%E2%80%9D`` instead of just http://www.astropy.org/astropy-tutorials/UVES_old.html - I think this is just the wrong kinds of ``"``

eteq:
This was discussed in-person, but I'm leaving this comment just so anyone looking at this issue in the future knows: there's a downside to this approach in that it doesn't re-direct search engines. The better thing to do is have the web server return a 301 redirect.  However, we don't know how (and maybe it's impossible?) to do this with Github pages.  So this style redirect is the next-best thing.

kelle:
@eteq This is ready for re-review. :)

adrn:
Cleaning out old PRs in anticipation of new front-end launch.

PR #228, opened 2018-05-04
Try circleci

adrn:
After a conversation with @cadair, he convinced me that we might be able to solve our build issues by running the tutorial build on CircleCI and then pushing the documentation from there to either readthedocs or github pages. The advantage of building on a CI service is that we can cache data files, which is what causes our timeouts and headaches on RTD right now. This is a PR to test out the CircleCI build - *do not merge* yet!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,21 +16,22 @@ pip-run: &pip-install
     pip install -q -r pip-requirements.txt
     pip install astropy_helpers sphinx
 
-add_ssh_keys: &add_ssh_keys
-    add_ssh_keys:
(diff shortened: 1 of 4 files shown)


adrn:
I manually killed the travis build. 

When I'm feeling particularly confident, I'll merge this and test the final bit of dangerous functionality: a force-push to a separate branch that contains the RST versions of the executed notebooks. RTD can then look for this branch, and in theory, won't have to execute the notebooks because the RST will already exist...

image

adrn:
Alright, I'll do the last bit of debugging live!

PR #238, opened 2018-06-19
Tutorials for installing Python and Astropy, conda environments, and developer version of Astropy

eblur:
Written tutorials (coming out of Python in Astronomy 2018) for:

01 - installing Python and Astropy with Anaconda

02 - Setting up and managing conda environments, and with Jupyter notebook

03 - Setting up a developer version of Astropy, including links to Github documentation

I welcome feedback and will request content reviewers.

Before merging, we should figure out where this should go on the Learn Astropy landing page. @MananAgarwal @adrn @eteq


tutorials/02-conda_envs.rst
```diff
@@ -0,0 +1,114 @@
...
+worry about version conflicts.
+
+Step 1: Set up a Python 2 environment
+-------------------------------------
+
+-  Create a new ``conda`` environment
+
+   ::
+
+       conda create -n python2 python=2.7 anaconda
```
mwcraig: I wonder if having them install anaconda is overkill. I typically create a new environment with the minimum required packages that I need (e.g. `conda create -n python2 python=2.7 astropy`). On the other hand I'm not the typical user...

tutorials/02-conda_envs.rst
```diff
@@ -0,0 +1,114 @@
...
+Step 1: Set up a Python 2 environment
+-------------------------------------
+
+-  Create a new ``conda`` environment
+
+   ::
+
+       conda create -n python2 python=2.7 anaconda 
+
+**NOTE:** ``anaconda`` at the end is optional...
```
mwcraig: Maybe add some like "omitting it will make the environment smaller, but you will need to install the specific packages you need in the environment after activating it."

tutorials/02-conda_envs.rst
```diff
@@ -0,0 +1,114 @@
...
+
+-  Activate the Python 2 environment and install any additional packages
+   you need to run your favorite code library. Here, we show you how to 
+   install the base version of Astropy. If you want to download
+   Astropy and **all** affiliated packages, you can install ``stsci``
```
mwcraig: What all is in `stsci`? 

If we want an astropy "official" metapackage for all of the affiliated packages and their dependencies we can construct that (fairly) painlessly.

tutorials/02-conda_envs.rst
```diff
@@ -0,0 +1,114 @@
...
+        source deactivate
+
+Step 3: Set up a Jupyter Notebook for the new environment
+---------------------------------------------------------
+
+-  Activate your custom Python 2 environment:
+
+   ::
+
+       source activate python2
```
mwcraig: I would add instructions to install both `notebook` and `ipykernel`. I've been bitten several times by activating an environment in which I forgot to install notebook, running `jupyter notebook` and getting a jupyter not in my environment. Or add notebook to the things to grep below.

tutorials/02-conda_envs.rst
```diff
@@ -0,0 +1,114 @@
...
+   ::
+
+       jupyter notebook
+
+-  When you click on *New*, you should see a drop down list of options
+   that include "python2", the environment display name we chose above.
+
+-  If you would like to change the environment for an existing Jupyter
+   Notebook, click on *Kernel*, then *Change kernel*, and select the
+   environment you would like to use.
```
mwcraig: Interesting; I didn't know that that worked! (Though I've encountered the pain of using a kernel that doesn't match my environment....)

tutorials/03-astropy_dev_env.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Step 3: Create a new conda environment for developing astropy
+-------------------------------------------------------------
+
+-  Inside the astropy-dev folder, create a new conda environment called
+   *astropy-dev* and activate it.
+
+   ::
+
+       conda create -n astropy-dev python=3 anaconda
```
mwcraig: I think we *don't* want anaconda here because that will include a conda install of astropy (which is included in the anaconda distribution). This is probably better:

```
conda create -n astropy-dev --only-deps python=3 astropy
```
That will create an environment with all of astropy's dependencies without installing astropy itself.

tutorials/03-astropy_dev_env.rst
```diff
@@ -0,0 +1,159 @@
...
+       source activate astropy-dev
+
+-  Inside the astropy-dev folder, install the Astropy from the source
+   code:
+
+   ::
+
+       python setup.py build
+       python setup.py install
```
mwcraig: I find `python setup.py develop` more useful in this scenario; otherwise, the install step needs to be rerun any time they change the code (if they want to see those changes in the environment they are in). That last bit might not apply when you run `python setup.py test`, which essentially creates a separate install in a temporary folder (I think).

tutorials/03-astropy_dev_env.rst
```diff
@@ -0,0 +1,159 @@
...
+       git pull upstream master
+
+Step 3: Push the changes to your fork on Github
+-----------------------------------------------
+
+-  Use ``git push`` to update your Github fork of Astropy:
+
+   ::
+
+       git push origin master
```
mwcraig: Add that they may need to do a force push if they have in the past made changes directly to master (which seems o be a pretty common occurrence).

tutorials/02-conda_envs.rst
pllim: `stsci` is a metapackage from STScI -- https://github.com/astroconda/astroconda-contrib/blob/master/stsci/meta.yaml

It also installs two other metapackages:
https://github.com/astroconda/astroconda-contrib/blob/master/stsci-hst/meta.yaml
https://github.com/astroconda/astroconda-contrib/blob/master/stsci-data-analysis/meta.yaml

tutorials/01-conda_astropy_install.rst
```diff
@@ -0,0 +1,139 @@
...
+-----------------------------------
+
+Many `Astropy affiliated
+packages <https://www.astropy.org/affiliated/>`__ can be found on
+*astroconda* channel, maintained by AURA and STScI. To add this channel
+to Anaconda's package search list, run the following command:
+
+::
+
+    conda config --add channels http://ssb.stsci.edu/astroconda
```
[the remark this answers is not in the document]
pllim: FWIW `astroconda` is targeted at STScI usage (HST, JWST, WFIRST). Its recipes available at https://github.com/astroconda/astroconda-contrib . I don't think it has all the affiliated packages for Astropy project.

tutorials/01-conda_astropy_install.rst
```diff
@@ -0,0 +1,139 @@
...
+`More information on managing channels in
+Anaconda <https://conda.io/docs/user-guide/tasks/manage-channels.html>`__
+is available on the main documentation pages.
+
+-  Here's an example for downloading a few commonly used Astropy
+   affiliated packages, directly from the *astroconda* channel:
+
+   ::
+
+       conda install -c astroconda astroquery photutils specutils
```
pllim: I don't think `-c astroconda` works. You need to provide the full URL.

p.s. I do agree `astroconda` is probably overkill if your focus is Astropy tutorials. We have https://github.com/spacetelescope/stak-notebooks etc for STScI specific things.

eteq:
@eblur - re: where it should go on the Learn page: I think maybe it should be listed as a standard tutorial but *also* be linked at the top of the tutorials page, and any future "guides" page?  That's what it's most relevant for, and I don't think a user who's at the Learn landing page needs these unless they actually make it to the tutorials or guides?

tutorials/01-conda_astropy_install.rst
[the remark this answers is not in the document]
bsipocz: I feel this tutorial should use the astropy channel rather than astroconda, given that for the latter we don't have much control (as I understand it also includes nightly builds for some packages rather than released versions?, etc) and the primary "official" place from the astropy POV is on the astropy channel

tutorials/01-conda_astropy_install.rst
bsipocz: for astroquery we've switched to continuous delivery, that currently means that the preferred way of install is 
`pip install --pre astroquery`. The conda version by definition won't be the latest release (we may change to that, but the machinery is not yet in motion for it)

tutorials/02-conda_envs.rst
bsipocz: I also wondered about this, does anaconda needed, or better start with miniconda for all cases? (I don't know as I don't use conda much beyond CI).

tutorials/03-astropy_dev_env.rst
```diff
@@ -0,0 +1,159 @@
...
+-  An account on `GitHub <https://github.com/>`__
+
+This tutorial will show you how to fork, clone, and install the
+development version of Astropy from the command line (terminal).
+
+Step 1: Fork the Astropy repo on Github
+---------------------------------------
+
+-  Log into your account on Github
+-  Go to https://github.com/astropy/astropy and click on **Fork** in the
```
bsipocz: I'm not sure, would screenshots be an overkill here?

tutorials/03-astropy_dev_env.rst
```diff
@@ -0,0 +1,159 @@
...
+Step 4: Make a new branch for editing the source code
+-----------------------------------------------------
+
+-  Inside the astropy-dev folder, create a new branch to work in. Try to
+   pick a descriptive name for your branch:
+
+   ::
+
+       git branch my-code-update
+       git checkout my-code-update
```
bsipocz: These can be done in one step, `git checkout -b my-code-update`, though the 2 steps above can be pedagogically more useful

tutorials/03-astropy_dev_env.rst
```diff
@@ -0,0 +1,159 @@
...
+
+Congratulations! You're ready to edit the astropy source code safely.
+Use ``git`` for committing your changes and pushing those changes to
+your ``my-code-update`` branch on Github.
```
bsipocz: add an example for the push here, too

tutorials/01-conda_astropy_install.rst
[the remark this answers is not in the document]
pllim: @bsipocz , nightly builds are not in `astroconda` itself but in another channel that we don't talk about (ala Fight Club). But I agree that `astroconda` is not something Astropy project has direct control over, as it is for STScI use.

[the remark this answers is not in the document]
eblur: I agree with using the astropy channel instead of astroconda. What is the link / your suggested install command?

`conda install --channel "astropy" package` ??

(found via https://anaconda.org/astropy/repo/installers)

tutorials/02-conda_envs.rst
eblur: For the purposes of this tutorial -- creating a separate Python 2 installation -- a new user will probably want *everything* they might possibly use in both installations. Developers, who likely have many environments, seem more likely to need/want a slimmer installation. I'll work on communicating this in the tutorial (i.e., that `astropy` and `miniconda` are options to replace `anaconda` in this line).  

**Can anyone link me to a handy resource guide on the differences between miniconda and anaconda?** (A Quick google search didn't help)

tutorials/03-astropy_dev_env.rst
eblur: Does `python setup.py develop` replace both the build and the install lines? Or just the install?

I've never used it! Thanks so much for the info.

tutorials/01-conda_astropy_install.rst
eblur: Should I use `-c astropy` instead? or nothing...

eblur: Mixing `conda install` with `pip` will confuse many new users. As long as the conda releases are stable and ongoing, let's stick to that.

People who need the latest versions on a regular basis are probably developing. How about adding the pip install line to the "Setting up a developer version of Astropy", instead?

tutorials/03-astropy_dev_env.rst
eblur: Not overkill. More like a desired upgrade, which probably won't happen for awhile.

tutorials/01-conda_astropy_install.rst
bsipocz: yes, please use `-c astropy` to install astropy affiliated packages. 
Only the core astropy package is available in the defaults channel, so if you have nothing these packages cannot be conda installed.

bsipocz: No, the idea here is that these are stable releases. For astroquery it isn't true that the latest "stable" is more reliable than these auto deployed releases. 
We'll eventually fix the version numbering, so normal pip install (without ``--pre``) will be able to pick up the releases, but I'm not sure we can do a continuous supply of conda packaging, too.

The developer version is a completely different question, one has that for having access to the git tree, local docs, etc.

bsipocz: also, users will eventually need pip even when using conda as not all packages are available on conda

tutorials/01-conda_astropy_install.rst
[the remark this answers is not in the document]
pllim: I thought @mwcraig moved stuff from `astropy` to `conda-forge`, no?

tutorials/02-conda_envs.rst
pllim: Does this help? https://stackoverflow.com/questions/45421163/anaconda-vs-miniconda

tutorials/03-astropy_dev_env.rst
mwcraig: It replaces both lines; it will build as necessary.

tutorials/02-conda_envs.rst
mwcraig: @eblur -- I don't know of a great online resource, but maybe this helps (and feel free to add it to the tutorial if you want):

## What is the difference between *miniconda* and *anaconda*?

**miniconda** is a fairly small installer that comes with python, conda and little else. One then uses `conda install...` to install any desired conda packages, or one uses `conda create ....` to create an environment with any packages you want.

**anaconda** is really two things in one: a python installer (like *miniconda*) and a conda *metepackage*. A metapackage is a list of packages and version numbers. Anaconda (the company) works hard to make sure that all of the package versions in anaconda (the metapackage) work well with each other. The full list of packages included with anaconda is at https://docs.anaconda.com/anaconda/packages/pkg-docs

## Which one should I choose?

Although it may be tempting to install *anaconda* because it comes with more packages, in the long run *miniconda* is more flexible and easier to maintain. Part of the motivation for this choice is that it encourages you to create *environments* regularly. That is particularly important if you want to have a separate environment for development and for everyday science use.

tutorials/01-conda_astropy_install.rst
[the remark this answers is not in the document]
eblur: Done

tutorials/02-conda_envs.rst
eblur: @mwcraig This text is super useful! Where did you get it? I would like to add verbatim, but feel morally obligated to cite the reference.

tutorials/01-conda_astropy_install.rst
eblur: Done

eblur: My solution here is to remove astroquery from this example. I've added notes at the end of the document telling users to use pip for managing astroquery.

tutorials/02-conda_envs.rst
eblur: Doing this.

tutorials/02-conda_envs.rst
eblur: adding a link to Anconda vs Miniconda on the conda readthedocs

tutorials/02-conda_envs.rst
eblur: I'm not going to include these details in the document.

tutorials/02-conda_envs.rst
eblur: If they install the anaconda distribution, this shouldn't be an issue, right? ;)

But I hear you, I've run into that problem sometimes, too. I'm adding instructions for installing ipykernel just in case. That's the problem I run into more frequently.

eblur: Oh wait ... those instructions are already there. Marking as resolved.

tutorials/03-astropy_dev_env.rst
eblur: Done, and added a note describing the only-deps flag

tutorials/03-astropy_dev_env.rst
eblur: Done

tutorials/03-astropy_dev_env.rst
eblur: I added a note

tutorials/03-astropy_dev_env.rst
eblur: I'm going to leave them separate.

tutorials/03-astropy_dev_env.rst
eblur: Done

bsipocz:
@eblur - This seems to be in need of a rebase as many commits have brought in from master, similarly as in https://github.com/astropy/astropy-tutorials/pull/227#issuecomment-434425826.
Ping me on slack if interested in doing a pair coding rebase for this.

adrn:
This was moved to #308

Issue #222, opened 2018-05-01
Dust Extinction Tutorial: unextinguish a spectrum


karllark:
Show how to use the dust_extinction package to unextinguish (deredden) a spectrum.

Use observed ultraviolet spectra from a FITS binary table file -> use astropy.io.fits.
Include photometry from Simbad (optical/NIR) -> use astroquery
Convert photometry from magnitudes to fluxes -> use astropy.units
Unextinguish the spectrum using a dust extinction model -> use dust_extinction package
Make a plot of the original and unextinguished spectrum -> use matplotlib
Write the unextinguished spectrum to a FITS binary table file -> use astropy.io.fits.

Mention that one of the goals of the the dust_extinction package is to provide the equivalent of the IDL ccm_unred.pro and fm_unred.pro. 

Possible bonus content.
Extinguish a model spectrum to predict and show the impact of dust extinction.  Could also include the non-linear impact of increasing extinction on broad band photometry.
Would need a stellar atmosphere model spectrum.  Could be a 2nd tutorial or a part of this one.  Advice needed.

kelle:
This is a great idea! Looking forward to the first draft. :)

Please take a look at the [Contributing](http://astropy-tutorials.readthedocs.io/en/latest/contributing.html#contributing-page) doc to help plan the tutorial.

I would start small and then let's discuss the 'bonus content' once you have the meat put together.

krislars:
I've been working on this.  Thanks to @emirkmo for getting this tutorial started, and to @karllark and @pllim for checking over my use of their packages.

There's lot more to be done, but here is a start.  If you are logged into a google account and open the file with colaboratory, you can click comment in the upper right or in any cell to add a comment:
https://drive.google.com/file/d/1zMgoGHMsWDnGLD-wtHUJheCNb2yDbqeh/view?usp=sharing

Here is a gist for just a quick look:
https://gist.github.com/krislars/44a203ff867f1578cc2248e4f890fa16

kelle:
Hi @krislars , would you mind moving your notebook to our new Astropy Learn Draft Tutorials folder on google drive? https://drive.google.com/drive/folders/1k1HnAss-pHTA3E5s9jB1OMphpBjQ-U7l
That should make it way easier for us to work collaboratively. I've sent an invite to your edu address.

PR #240, opened 2018-06-29
Adding color-excess tutorial

krislars:
This new tutorial uses the affiliated packages dust_extinction (@karllark) and synphot (@pllim) to deredden a spectrum and "convert" an extinction curve into magnitudes.  This is my first pull request on this project.

### DIFF
--- a/tutorials/notebooks/quantities/quantities.ipynb
+++ b/tutorials/notebooks/quantities/quantities.ipynb
@@ -13,7 +13,7 @@
     "* TODO\n",
     "\n",
     "## Keywords\n",
-    "units, matplotlib, radio astronomy\n",
+    "units, matplotlib, radio astronomy, data cubes\n",
     "\n",
     "## Summary\n",
     "\n",


karllark:
It would be great if optical/NIR photometry for rho oph could be included in the "Example 2."  This would make a more direct connection to "Example 3" that focuses on optical/NIR photometry.  Maybe even with a query to Simbad for the photometry using astroquery?

karllark:
Related to issue #222.

krislars:
Good addition, @karllark!  I'll work on that while I wait to hear from maintainers about the travis tests.

adrn:
Thanks for this! Yea, any new dependencies also need to be added to these two files:
* https://github.com/astropy/astropy-tutorials/blob/master/pip-requirements.txt
* https://github.com/astropy/astropy-tutorials/blob/master/environment.yml

krislars:
Yippee, the tests pass!  However, I did get "package not found" errors from Travis even with the dependencies added to those two files.  The only way I could get the tests to pass is to include `!pip install dust_extinction` and `!pip install synphot` lines in the tutorial.

@adrn , can you advise me on a better way?  I suspect that I'm having trouble because these are affiliated packages not on the astropy conda channel.

krislars:
I've updated the tutorial with astroquery to obtain the IUE spectrum from MAST and optical photometry from Simbad.  Thanks for the idea, @karllark!

adrn:
Will take a look at this ASAP - thanks for the patience!

eblur:
This tutorial is a great idea and super useful! Thank you so much for contributing it.

I'm about to attach a long review. 80% of it is nit-picky stuff around word choices. Don't feel obligated to take every suggested change at face value. I encourage you to make the desired clarifications in your own voice.

The major changes I am requesting are:

* Adding a "Companion Content" portion to the preamble text to list the links and resources that are weaved throughout the tutorial

* Adding 1-2 items to the "Learning Goals" portion around Astropy Quantities and visualization

* Clarifying how to use `quantity_support`

eblur:
Okay -- here's the long review. Not for everyone: I had to use nbviewer online because Github was having trouble rendering it.
http://nbviewer.jupyter.org/github/astropy/astropy-tutorials/blob/eddd09a7dbef817b14e57698e37d9f5d1a29a716/tutorials/notebooks/color-excess/color-excess.ipynb

## Preamble text / imports:

(1) The pip install commands -- Move the instructions for pip install into the pre-amble (header) text. Users are likely to download and run these notebooks as is, and this could potentially cause problems for their system. See https://astropy-tutorials.readthedocs.io/en/latest/rst-tutorials/FITS-cubes.html for an example of a notebook that uses packages that might not be part of the user's default installation. (Note that there is supposed to be bulleted text, and the notebook rendering is just off a bit)

(2) Add "## Companion Content" to the preamble with a link to the dust_extinction documentation.
https://astropy-tutorials.readthedocs.io/en/latest/contributing.html

(3) I notice that this notebook imports individual classes and functions from the desired code libraries. This might add a little confusion for users, because they have to go back to the imports to see what package they came from. I have a preferences for importing the whole package namespace (e.g., `import dust_extinction.dust_extinction as dext`, then use `[dext.CCM89,dext.F99]` in cell 4; or `from astroquery import mast`, then use `mast.Observations` in cell 5). Any thoughts from @pllim, @kelle , @adrn ?

## Example 1:

(1) "What variation there is is" --> "Any variation is" or "The little variation that exists is" (The original text, while technically fine, is a little awkward)

(2) "extinction to reddening in the blue-visual" --> "extinction (A_V) to reddening in the blue-visual (E_{B-V})"

(3) "R_V =" --> "R_V \equiv" because it's a definition

(4) "see the package documentation for more details" --> Might consider removing this (after adding link to docs in pre-amble). Or turn this into a link to the documentation again.

(5) "grey or wavelength-independent extinction" --> "wavelength-independent, or grey, extinction" (The first phrasing makes it sounds like they are two mutually exclusive things. Adding the commas indicates a re-phrasing the original concept.)

## Example 2:

(1) "with optical flux of the star from Simbad photometry" --> "using information about the optical flux of the star obtained from Simbad photometry archives" Is that accurate? The optical photometry is not dereddening the spectrum, the models are.

(2) "First, we fetch the spectrum from the IUE archive" --> "First, we will use astroquery to fetch the archival IUE spectrum from MAST"

(3) Link to MAST archive or information about MAST archive. Also information about IUE...

(4) "t_wlr" --> "print(t_wlr)"

(5) "Quantities attach units to the values" --> This is kind of vague. How about, "The `.quantity` extension will keep the units of the Table column attached to the numpy array values."  Something like that. Could also link to documentation on quantities in astropy.units.  This seems like something special to the Table class.

(6) Can you add a one sentence description or definition for "zero-point flux"? I'm an observer but I don't work with optical photometry.

(7) Link to the ned webpages in the text instead of the code comments.

(8) "However, we notice that the zero-point fluxes ..." --> could be "However, the zero-point fluxes..."

(9) Please add "Converting from frequency to wavelength with astropy.unit equivalencies" to the learning goals.

(10) "Now it easy to convert from magnitude to flux:" --> "Now we can convert from magnitude to flux by doing the following:"

(11) "We went though the trouble to get the quantities correct so that ..." -- I think these comments are a little too informal for a tutorial. I recommend omitting it. Another option is to show give an example for the user to run that will intentional return an error. However, that requires extra  work including editing the JSON text in the ipynb file.

(12) It is relatively unclear how `quantity_support` works from this example. It is not used during the plot commands. It is not clear what the call to `quantity_support()` does. The import should be at the top. Please add `astropy.visualization.quantity_support` in some form to the learning goals in the preamble text. Please add a sentence or two explaining how `quantity_support` is being called or used in this example.

(13) "A-ha! Notice that quantity_support converted wavelengths into the same units" --> `quantity_support` was not called during the plotting commands, so this is not clear

(14) "as an example of a large-R_V" --> could be "for having a large-R_V"

(15) "Notice that the absorption feature at 2175 Angstrom is removed." --> "Notice that, by dereddening the spectrum, the absorption feature at 2175 Angstrom is removed" or "Notice that the absorption feature at 2175 Angstrom is removed in our dereddened spectrum" ...

## Example 3:

(1) Please add Bessell & Murphy (2012) to the list of companion material in the preamble text. Is it a published paper? If so, please add an ADS link.

(2) consider replacing "rigorous" with "complex"

(3) Move the synphot.config import to the top. Consider importing just synphot and calling synphot.config in this cell, if possible.

(4) "The lines above ..." is sort of unclear. Do you mean "The SpectralElement.from_filter call gets filter transmission curves from an FTP server at STScI."  ?

(5) Put the text about the FTP server and optional lines before the code that uses them.

(6) Add the link to the synphot documentation to the companion materials in the preamble.

(7) "Hey, there's that ..." --> "Note that" ... this is a matter of personal preference about the tone. But I enjoy the enthusiasm! Yay dust! ;)

(8) "Synthetic photometry means sending the astronomical flux through a certain filter:" --> this intro into the next code chunk felt awkward. I recommend writing out the whole sentence. The comment is also pretty clear. So maybe you can combine everything into the following statement before the cell -- "Synthetic photometry refers to modeling an observation of a star by sending the theoretical model for the astronomical flux through a certain filter, then integrating:"

(9) "The 'effective stimulus' in synphot refers to the integrated flux ..." --> "The `effstim` method in `synphot` refers to the 'effective stimulus', which is the integrated flux ..." (It took me awhile to find where "effective stimulus" showed up in the code chunk.)

(10) "Invoking the Vega flux unit will ..." --> "Setting `flux_unit='vegamag'` will ...."

(11) "We're feeling ready..." --> I recommend "We're now ready" or "Now we're ready"

(12) "Handy for plotting!" --> I recommend removing this line.

## Exercise:

(1) I recommend changing the last line so that it's more of a note, rather than sounding like an answer to the original question.  I recommend:
"Note that the photometric extinction changes because the filter transmission is not uniform. The throughput of the filter depends on the shape of the background source flux."

Is "The observed throughput from the filter" slightly more accurate than saying "The throughput of the filter"? The second way just sounds a bit like the filter changes based on the source.

pllim:
Maybe also a brief discussion along the line of karllark/dust_extinction#47 ? I see that this tutorial only used `dust_extinction` and not `dust_attenuation`.

krislars:
Thanks very much for comments, both the broad and the detailed.  I will take a look at comments from @eblur and @pllim and make changes in the next two weeks.

eblur:
Let me know how it's going @krislars 

One option is for us to edit it together in Colaboratory. Let me know if you would like to do that.

krislars:
Thanks for the check-in, @eblur.  It would be great to work on it together in Colaboratory.  My day job is kicking my butt right now, and I could really use the help on this tutorial.

eblur:
Totally understandable, @krislars 

I put your latest version of the tutorial on a Google drive, which can be opened with Colaboratory.
https://drive.google.com/open?id=1f700EQ0joHLS3fGbDusQ9jUVY3EJthn9

I've only edited through Section 1. Will come back and edit more later.

eblur:
I have updated through Example 2 on Colab

krislars:
I have updated through Example 3.  The version on Colaboratory is the updated version.  Most of the changes suggested by @eblur have now been implemented in one way or another.  Many thanks to @eblur for making this tutorial so much better and for keeping the edits going!  I'm really glad I got to meet you at AAS.

eblur:
Oops, I was working on pushing to this branch. I didn't know that a force push was going to close the PR.

No matter, we needed to start a new branch with a clean git history. So I will start that now.

PR #244, opened 2018-07-27
Allow kernel setting from environment variable

eteq:
This adds a bit to the ``conf.py`` to grab an environment variable to use for determining which kernel runs the notebooks.  Also adds a bit to the dev docs to explain this.

I'm not super happy using an environment variable for this, but I couldn't come up with any other way that works with ``make html``.  Open to ideas but if no one else has something this is better than nothing.

cc @MananAgarwal @adrn


adrn:
Seems fine to me!

PR #253, opened 2018-08-21
some minor fixes to coordinates intro

eteq:
I *think* this fixes a rendering problems when there's a backtick inside a link that make the coordinates intro look fine as a notebook but not on the web site.

I'm not 100% sure this will fix the rendering problem... it's *possible* we need to do some nbconvert hack to make it work right, but we shall see.

At the same time it adds a file that was missing to the coordinates intro, since that's worth it regardless.


adrn:
I still can't get @cadair's Giles to work, so I'll do it manually for now...

Here's the built documentation for this PR: https://circleci.com/gh/astropy/astropy-tutorials/111/artifacts/0//home/circleci/project/build/html/index.html

eblur:
Hi @eteq sorry for this but due to recent merge of copy edits (PR #262) there are now conflicts that need to be resolved. Can you clean that up in the next week?

eteq:
@eblur - I rebased and I think this is now good?

PR #287, opened 2018-12-05
[WIP] Back to pushing to gh-pages

adrn:
Just so we have something working with the new design (I still can't figure out the JS issue!), here is a test of pushing the built website to gh-pages.


eteq:
FYI: this will require me to change the subdomain redirect to make "live".  And that can take 24 hr to a week to propogate across the internet.  So if we have an RTD fix coming soon, might not want to do that.

adrn:
See #294 instead

PR #293, opened 2018-12-07
Some cube tutorial fixes

keflavich:
Fairly minor things, but it's quite a bit simpler this way

(also, you should pretty much never do `cube.hdu.data[...]`, there are always better ways, unless we've made a huge design mistake)


adrn:
Thanks @keflavich! We moved this tutorial into a "quarantine" until we decide what to do with the data volume. So I rebased this and made a slight grammar tweak.

PR #306, opened 2018-12-18
update nb to 5.7.2

eteq:
Currently github complains of a security vulnerability in versions of the notebook < 5.7.2 .  This PR just updates the various places we give an nb version to address that.

cc @adrn

### DIFF
--- a/.rtd-environment.yml
+++ b/.rtd-environment.yml
@@ -13,6 +13,7 @@ dependencies:
   - scipy=1.0 # needed for coordinates cross-matching
   - jupyter=1.0
   - notebook>=5.7.2
+  - sphinx-astropy
   - aplpy=1.1.1
   - spectral-cube=0.4
   - reproject=0.3.2
--- a/.rtd-environment.yml
+++ b/.rtd-environment.yml
@@ -12,7 +12,7 @@ dependencies:
   - numpy=1.14
   - scipy=1.0 # needed for coordinates cross-matching
   - jupyter=1.0
-  - notebook=5.4
+  - notebook>=5.7.2
   - aplpy=1.1.1
   - spectral-cube=0.4
   - reproject=0.3.2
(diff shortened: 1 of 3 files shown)


adrn:
Why is the circleCI build failing? It's complaining about astropy helpers but I don't know what is trying to import the helpers...I thought I removed all references to the helpers. Did you pull from upstream before making this branch?

pllim:
I think you need to add `sphinx-astropy` to your dependency. I saw a similar error on RTD for my other projects after upgrading to helpers 3.1.

.rtd-environment.yml
```diff
@@ -12,7 +12,7 @@ dependencies:
   - numpy=1.14
```
pllim: The indentation here bothers me.

.rtd-environment.yml
```diff
@@ -12,7 +12,7 @@ dependencies:
...
   - scipy=1.0 # needed for coordinates cross-matching
   - jupyter=1.0
-  - notebook=5.4
+  - notebook>=5.7.2
```
pllim:
```suggestion
  - notebook>=5.7.2
  - sphinx-astropy
```

pllim: p.s. Well, the failure is at CircleCI, so probably need to update the other files as well but you get the idea.

p.p.s. Albeit this being an educated guess, I could be wrong. Only one way to find out!

adrn: This is RTD - we actually don't use RTD anymore, and sphinx-astropy is in the circleci config

adrn:
I just checked - it looks like this branch is based on an old version of master.

.rtd-environment.yml
pllim: Well... Ignore me then... :sweat_smile:

adrn:
I rebased and now circleci is passing.

Issue #320, opened 2019-01-25
"Using scipy.integrate" tutorial has some weirdness in the display


ceb8:
"[Using scipy.integrate](http://www.astropy.org/astropy-tutorials/rst-tutorials/units-and-integration.html)" has a few problems that show up in the website display, but that look fine in the actual notebook:
- The cell that contains imports in the [summary section](http://www.astropy.org/astropy-tutorials/rst-tutorials/units-and-integration.html?highlight=filtertutorials#summary) is not displaying at all
- The links in the [How to simulate actual observations](http://www.astropy.org/astropy-tutorials/rst-tutorials/units-and-integration.html#how-to-simulate-actual-observations) section are showing their raw markdown instead of looking like proper links (Also in the [Integrating Using Gaussian Quadrature](http://www.astropy.org/astropy-tutorials/rst-tutorials/units-and-integration.html#integrating-using-gaussian-quadrature) section)
- In the [Challenge Problems](http://www.astropy.org/astropy-tutorials/rst-tutorials/units-and-integration.html#challenge-problems) section some of the latex is rendered but some shows up as raw latex
- The "Download Tutorial Notebook" and "Interactive Tutorial" buttons at the top of the notebook work, but the ones at the bottom do not.

kakirastern:
Hi @ceb8, 
Yes, I see the problems showing up in my web browser ("Chrome") too, but they indeed don't show up in the actual downloaded notebook. Will follow up on the Issue by updating the notebook in a new PR soon, if that fixes that problems...

kakirastern:
Opened PR #330 to deal with issues raised here.

kakirastern:
Have dealt with the following issues: 
> * The cell that contains imports in the [summary section](http://www.astropy.org/astropy-tutorials/rst-tutorials/units-and-integration.html?highlight=filtertutorials#summary) is not displaying at all
> * The links in the [How to simulate actual observations](http://www.astropy.org/astropy-tutorials/rst-tutorials/units-and-integration.html#how-to-simulate-actual-observations) section are showing their raw markdown instead of looking like proper links (Also in the [Integrating Using Gaussian Quadrature](http://www.astropy.org/astropy-tutorials/rst-tutorials/units-and-integration.html#integrating-using-gaussian-quadrature) section)
[...]

PR #330, opened 2019-02-18
Fixing bugs in units-and-integration notebook

kakirastern:
Addresses Issue #320 to correct for browser behavior of the scipy.integrate tutorial.


kakirastern:
Think the PR is ready for a review...

PR #339, opened 2019-02-26
Fix broken logo link now that pathto option adds .html extension

kakirastern:
To correct link to source image file in `_static` path. Previously used `pathto(file, 1)`. Now switching to `pathto(document)` to test behavior.


kakirastern:
Problem seems to have fixed... I checked the website a bit prematurely.

PR #340, opened 2019-02-26
Attempt to fix 1st-time-load-tutorials-in-window with "no tutorial found" bug

### LINKED ISSUE #337: Sometimes the tutorials don't load when going to main tutorials page
About 50% of the time, when I go to learn.astropy.org and click "tutorials," the page says "No tutorials found!" until I refresh, and then the tutorials appear. I know @eblur and @kelle have run in to this issue as well. Do you have any idea what could be causing this? I think this is the highest priority bug to work on in the near term!

kakirastern:
Fixes #337. 

Attempt to fix "no tutorial found" bug when loading tutorials after entering Learn Astropy site for the first time in a browser window. This behavior is gone if staying in the same browser window and starting to search again.

### DIFF
--- a/tutorials/themes/tutorials-theme/tutorialfilters.html
+++ b/tutorials/themes/tutorials-theme/tutorialfilters.html
@@ -9,18 +9,18 @@
       var activeFilters = $.getQueryParameters();
       var filters = document.getElementsByClassName('custom-control-input');
       for(var f=0; filters[f]; ++f){
-          if(activeFilters.q[0].search(filters[f].value)!=-1)
-              filters[f].checked = true;   
+          if (activeFilters.q[0].search(filters[f].value)!=-1)
+              filters[f].checked = true;
       }
       document.getElementById("search-bar-input").value = '';
   });
   
   //no results found
-  $(window).on("load", function(){
-          if(!document.getElementsByClassName('result-card').length)
-              {document.getElementById('search-results').innerHTML += '<h3>No tutorials found!</h3>'}
-  });  
-
+  //$(window).on('load', function(){
+        //if(!document.getElementsByClassName('result-card').length)
+            //{document.getElementById('search-results').innerHTML = '<h3>No tutorials found!</h3>' ;}
+  //});
+  
   //update filters
   function applyFilter(){
       sessionStorage.scrollTop = $(window).scrollTop(); //to store the current scroll position
@@ -49,6 +49,7 @@
             filterDiv[i].style.display = "none";
           }
         }
+<<<<<<< HEAD
 
         var toggleCollapse, viewButton, j;
         toggleCollapse = document.getElementsByClassName("collapse");
@@ -68,6 +69,9 @@
         }
 
       }
+=======
+      }  
+>>>>>>> 15c6212... This is a combination of 20 commits.
 </script>
 
 <form action="">
--- a/tutorials/themes/tutorials-theme/tutorialfilters.html
+++ b/tutorials/themes/tutorials-theme/tutorialfilters.html
@@ -18,7 +18,7 @@
   //no results found
   //$(window).on("load", function(){
           //if(!document.getElementsByClassName('result-card').length)
-              //{document.getElementById('search-results').innerHTML = '<h3>No tutorials found!</h3>'}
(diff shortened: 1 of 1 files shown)


kakirastern:
Seems like something in the file "tutorials/themes/tutorials-theme/tutorialfilters.html" is causing the "search form" to automatically load instead of load on "submit", at least it appears that way for now...

kakirastern:
Hi @adrn and @MananAgarwal! I think I have the bug fixed by changing `.on("load", ...)` to `.on("submit"), ...)`, but am not sure if I have broken anything else in the process. On giles everything looks okay though. And, no errors were thrown while I build the html docs locally. Please check to see if I have made any mistake in the process of debugging. 

Thanks!

adrn:
Great! I'll take a look ASAP. Thanks

kakirastern:
@adrn You are welcome!

MananAgarwal:
I don't think this actually solves the issue as $windows.on('submit') won't ever get executed. So this effectively just removes this chunk of code. The result is that "No tutorials found" won't be displayed when there are actually no tutorials.

But maybe, for the time being, we can prefer this behaviour, i.e. not giving feedback when no tutorials are found, over the existing one, where sometimes "No tutorials found" is displayed even when there are tutorials.

kakirastern:
I just checked again with giles, and found that if after no matching results is found we get the following screen:

screenshot 2019-03-04 04 48 02

So, for this PR if the search field is empty after pressing submit the warning "_Your search did not match any documents. Please make sure that all words are spelled correctly and that you've selected enough categories._" is displayed. 

I can double-check to see what happens if I remove the `$windows.on('submit')` bit of code. And experiment further to change the warning to "_Select a filter or
reload this page to display tutorials._" if that is preferred.

kakirastern:
And for this PR for now, if no matching term is found, the warning "_Check the relevant keywords in the sidebox._" is issues instead.

kakirastern:
In hindsight I second @MananAgarwal's opinion that it is better without the previous behavior... So I changed the `+=` bit to `=` in the now `$window.on("load") part of the code to induce the same effect.

kakirastern:
That's not working... so will need to figure out something else. Let me try a bit more.

kakirastern:
I now have a sense that we need to add another conditional to the $window.on("load") if statements to test when we have both non-zero-length search-bar-inputs and zero-length result-card outputs... Will see if that works.

kakirastern:
I will need to find a way to set the `window.on("load")`'s `if` statement to `True` when:

1. The search results return nothing (which is the sole condition now), and one of
2. Search has been done by the upper search field or 3. Search has been done by the left hand key word field

Any suggestions to the possible conditionals I may need to use, @MananAgarwal?

kakirastern:
I think that "No tutorials found!" warning should apply only to the sidebar filter search from the design of things...

kakirastern:
So I have gotten rid of the `window.on(load)` code snippets entirely, which should fix the concerned bug for now. Will need to figure out a way to warn of no filter search is found in the case where only filters are selected, that might mean another PR and some changes to the file [searchtools.js](https://github.com/astropy/astropy-tutorials/blob/master/tutorials/themes/tutorials-theme/static/searchtools.js) instead... (Just an idea for now, may be other options out there.)

mwcraig:
From telecon on 3/5: problem is perhaps due to latency/a delay in the initial load of search results.

kakirastern:
Okay, then I will attempt a fix with callback(s) after init to make sure the the window.on(load) warning is issued after the ajax call.

kakirastern:
From my experimentation with this PR and PR #344 , it is better to merge the commits here in this PR then to try and correct some underlying latency/code-design issues as in #344 . This is because the warning "No tutorials found!" would show up along side all the available tutorial results in some browsers, like the following on Safari, which is quite undesirable: 

screenshot 2019-03-06 13 54 15

adrn:
Hi @kakirastern - As far as I can tell, the only commits in this PR comment out the "No tutorials found" text, it doesn't fix the issue. Let's discuss over email, and open a new PR when we've converged on a possible solution. Thanks!

kakirastern:
Hi @adrn! Sure, I will work on this further in Issue #344 and contact you via email as I work through a solution soon.

PR #341, opened 2019-02-26
Update old .ipynb tutorial notebooks

### LINKED ISSUE #335: Add TODO list items to Quantities tutorial
The TODO list of the Quantities tutorial at [http://www.astropy.org/astropy-tutorials/rst-tutorials/quantities.html?highlight=filtertutorials](http://www.astropy.org/astropy-tutorials/rst-tutorials/quantities.html?highlight=filtertutorials) now is filled by a placeholder. Will add TODO list items to it to complete the tutorial.

kakirastern:
Update some old .ipynb tutorial notebooks.  

Fixes #335.


kakirastern:
Think I might need to turn off caching of downloaded files for the tutorials, and then add the comment that `cache = False` can be modified to `cache = True` if we do not want to see the timeout or URLErrors. Will need to go through the tutorials one-by-one for that.

adrn:
I like the sentiment of the title, but I'm a little confused about the changes you've made: can you explain a bit? e.g., I don't think changing "HCG 7" to "hcg7" will fix the broader issues. I do, however, agree that we should review each notebook, identify places where they try to talk to the internet, and decide whether we want to keep those instances or replace them with cached values. I could imagine, for example, replacing:

```python
hcg7_center = SkyCoord.from_name('hcg7')
```

with something like:
```python
# This line accesses the internet - if you have an internet connection, try this out, 
# otherwise use the default cached values included below this line:
# hcg7_center = SkyCoord.from_name('hcg7')
hcg7_center = SkyCoord(ra=..., dec=...)
```

kakirastern:
Oh yeah, I originally thought it shouldn't make a difference between the queries "HCG 7" and "hcg7" too, but after testing a bit I found that somehow Travis CI thinks it does... (Maybe a long time ago there used to be no difference?) If I don't change this occasionally an error would be thrown due to the syntax. I will add the few lines of explanation you gave in https://github.com/astropy/astropy-tutorials/pull/341#issuecomment-468478448 however, to clarify the situation a bit more. I will test it again locally as well to confirm.

adrn:
@kakirastern I think the "HCG 7" issue is a red herring: try it locally, you'll see that the name "HCG 7" resolves just fine

With astropy v3.1:
```python
>>> from astropy.coordinates import SkyCoord
>>> SkyCoord.from_name("HCG 7")
<SkyCoord (ICRS): (ra, dec) in deg
    (9.81625, 0.88805556)>
```

I'm going to have a go at making some changes to solve #348, so this might not get merged in the end, but I'll keep this open for now. Thanks!

kakirastern:
Yup, no problem @adrn... Have updated https://github.com/astropy/astropy-tutorials/pull/341#issue-256423546 accordingly.

kakirastern:
> I think the "HCG 7" issue is a red herring: try it locally, you'll see that the name "HCG 7" resolves just fine

Yes, it is strange. I did try it locally and I remember the term "HCG 7" did resolves just fine... Not sure why Circle CI is getting so sensitive.

eblur:
I'm closing this PR because the original issue was handled in PR #370

Issue #302, opened 2018-12-10
New tutorial: computing Galactic orbits with Gaia + gala


adrn:
I wanted to try out the new colab workflow, so here is a new tutorial that highlights the affiliated package `gala` to do some orbit integration with Gaia data:
https://colab.research.google.com/drive/1DoU4hyLVbpC7_ms8vWGcC5lrKy61_Fbc

eblur:
@rkiman can you please review the content for clarity? :)

Some immediate observations:
* Learning Goals
* An exercise at the end

adrn:
Feel free to edit the colab notebook directly!

adrn:
Note: I added learning goals, added two exercises and addressed the comments @kelle left in the colab notebook. Does someone else want to take a look at the content? @rkiman ?

kakirastern:
Hi @adrn, is it okay for me to help edit/proofread the content if desired?

adrn:
@kakirastern sure! Please leave comments rather than editing the content directly.

kakirastern:
@adrn Cool, will proofread and leave comments where appropriate within the next few days.

adrn:
I think I've addressed all of @rkiman's comments! Any other comments?

kakirastern:
@adrn Give me say one more day to go over the tutorial... Will get back to you soon

kakirastern:
Sorry about the delay... I have reviewed the tutorial and left some comments in Colab. Hope they are useful in shaping the tutorial further.

adrn:
Thanks - these were useful comments. Thanks everyone for reading and commenting - I'm happy with this tutorial now, and will move it from Colab into an official PR.

adrn:
(Please leave further comments in #342)

PR #342, opened 2019-03-02
New tutorial: Galactic orbits with Gaia + gala

adrn:
After reviews on colab (see #302), I'm happy with the tutorial and am now moving it to an official PR.

@lglattly - this is ready to be copy-edited.


kelle:
Just curious, why not copyedit in the colab notebook? Seems like that would be easier for @lglattly.

adrn:
Oh duh, you're right, that would make more sense!

adrn:
@lglattly copy-edited the notebook while in colab, and I added the new notebook here.

PR #344, opened 2019-03-05
Try and add filter search warning when no results is returned

### LINKED ISSUE #337: Sometimes the tutorials don't load when going to main tutorials page

kakirastern:
To follow up on PR #340. To try and find a way to add back filter search warning when no tutorials are returned on a filter search. 

Fixes #337.

### DIFF
--- a/tutorials/themes/tutorials-theme/static/searchtools.js
+++ b/tutorials/themes/tutorials-theme/static/searchtools.js
@@ -9,6 +9,9 @@
  *
  */
 
+// Set ajaxSetup to synchronous to make a synchronous $.get call (but could be bad form).
+// See https://stackoverflow.com/a/25489055/9959073 for details.
+jQuery.ajaxSetup({async:false});
 
 /* Non-minified version JS is _stemmer.js if file is provided */ 
 /**
@@ -319,10 +322,6 @@ function splitQuery(query) {
     }
     return result;
 }
-
-// Set ajaxSetup to synchronous to make a synchronous $.get call (but could be bad form).
-// See https://stackoverflow.com/a/25489055/9959073 for details.
-jQuery.ajaxSetup({async:false});
 
 var docResultsCount = -1;
 /**
--- a/tutorials/themes/tutorials-theme/static/searchtools.js
+++ b/tutorials/themes/tutorials-theme/static/searchtools.js
@@ -346,7 +346,7 @@ var Search = {
               if (textstatus != "success") {
                 document.getElementById("searchindexloader").src = url;
               }
-            }});
+            }, async: false});
   },
 
   setIndex : function(index) {
@@ -577,7 +577,7 @@ var Search = {
                     listItem.slideDown(5, function() {
                       displayNextItem();
                     });
-                  }});
+                  }, async: false});
         } else {
           // no source available, just display title
           Search.output.append(listItem);
--- a/tutorials/themes/tutorials-theme/static/searchtools.js
+++ b/tutorials/themes/tutorials-theme/static/searchtools.js
@@ -320,6 +320,10 @@ function splitQuery(query) {
     return result;
 }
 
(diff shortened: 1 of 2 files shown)


kakirastern:
Testing if `$.ajax()`-related latency issues are the root cause of the 1st-time-load-tutorials bug...

kakirastern:
Yes, @MananAgarwal was right, it is indeed a latency issue. I have corrected for it by adding the option `async: false` to the two instances of JQuery `$.ajax()`in the file "searchtools.js". Think it is fixed now. Please check @MananAgarwal and @adrn. Thanks!

kakirastern:
Wait... seems like the same problem is cropping up on my other browser. Let me double check.

kakirastern:
I discovered the behavior is browser dependent too. Will need to work on it more before we meet next Tuesday and see if there is a way around it...

kakirastern:
So for my Safari, I see this on my screen:

screenshot 2019-03-06 13 54 15

I don't get this with Chrome, which is more like:

screenshot 2019-03-06 13 56 35

kakirastern:
I can confirm that the fix works only for Chrome, but not Firefox nor Safari... 🙁

kakirastern:
But I reckon there must be another way to do this... Will need to experiment and investigate some more, but it will take time

kakirastern:
@MananAgarwal I found this post and think it might be relevant for finding a solution to our current problem... [https://stackoverflow.com/a/28862458/9959073](https://stackoverflow.com/a/28862458/9959073). So a potential solution would be implement something like
```
if(!tbody.find('td').hasClass(settings.highlightClass)) {
    $('body').find('#empty').text('No results').show();
} else {
    $('body').find('#empty').hide();
}
```
together with
```
...
</table>
<p id="empty"></p>
</body>
```
Except in their case it is for jQuery.FilterTable to return an alert when no results found... I am not sure if it will work for the current code base though. What do you think?

kakirastern:
Or this one here might work: [https://stackoverflow.com/a/14340948/9959073](https://stackoverflow.com/a/14340948/9959073)...

kakirastern:
I just attempted another fix so that the code now works for both Google Chrome and Mozilla Firefox browsers, and the 1st-time-load error no longer bugs these two kinds of browsers. What I did was to add `jQuery.ajaxSetup({async:false});` to the very beginning of the file "tutorials/themes/tutorials-theme/static/searchtools.js", to make sure that ajaxSetup is set to synchronous to make all $.get and $.ajax calls synchronous, which could be in bad form in some other context for some other JS file. 

I am not sure what is going wrong with the Apple Safari browser... Maybe if we merge this code a warning about this sit being optimized for both Chrome and Firefox would be necessary.

MananAgarwal:
Hey @kakirastern, it is working for me on chrome, so I think we are on the right track to figuring this out. I am a little preoccupied with my midterm exams. I'll look into this properly on the weekend.

kakirastern:
No worries @MananAgarwal! I have checked the the latest solution sometimes work on my Firefox browser, and sometimes it does not... Anyway, we can discuss later. All the best with your midterm exams!

adrn:
Ah, great! This works for me locally too. Why do you say it doesn't work on Firefox or Safari? It seems to work for me on all browsers.

adrn:
Thanks for this - given that it solves the issue for at least Chrome and Safari, I'm going to merge and we can discuss whether or not it works for all browsers over in #337

kakirastern:
Hi @adrn! It's weird. I checked the `giles` version via incognito mode using both Firefox and Safari and it wasn't working. Let me check again now the PR's been merged. Wait... (may take a while).

Issue #348, opened 2019-03-07
Comment out and replace any small http requests


eblur:
e.g., SkyCoords from_name calls Sasame. Frequently breaks our testing infrastructure.

@adrn

kakirastern:
Yeah, I am fixing them in PR #341. Let me work on the PR some more before our meeting this coming Tuesday so they can be merged... Sorry about the delay @eblur!

kakirastern:
I will help @adrn to work on it... since I have already started

kakirastern:
Think the bulk of the work has been done in #341... But remember to rebase after every merge, I noticed that some notebook revert back to an older state after merge without rebase, such as the "units_and_integration" tutorial. 😉

The error was probably due to the spacing in "HCG 7". After you've changed that to something like "hcg7" the Circle CI error goes away. So, it is more like some syntax issue.

kakirastern:
Wait... why are the latest merges not showing up in "learn.astropy.org"? We might have another issue to deal with here. I just checked. Something isn't quite right. @adrn Can you confirm?

kakirastern:
I am guessing the public URL keeps changing as we merge pull requests, so we might need to update that manually regularly, or find means to automate the process. I have been using this [https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html) as a reference...

adrn:
@kakirastern Please try to keep discussion in an issue relevant to that issue - if you want to discuss something else, please just open a new issue. The circleCI build for master (that generates the live version of the site at learn.astropy.org) failed because of a transient issue (`skyserver.sdss.org` was down and one of the tutorials does a web request). But that failure just bolsters this particular issue...

I'm working on this locally, so I'll have a PR in soon.

kakirastern:
@adrn Yup, no problem... I will leave it for you to fix. 😉

kakirastern:
Sorry about the confusion earlier @adrn.

adrn:
No worries!

PR #350, opened 2019-03-17
[WIP] New way to skip cells that require remote connections

adrn:
In thinking about how to solve #348, I was brainstorming ways to remove any remote HTTP requests from tutorials on CI. We discussed:
1. Comment out lines that do web requests and include lines below the comments that read from cached files
2. (long term) use the STScI machinery that @eteq eluded to, that would automatically capture remote requests and serve cached files

But it occurred to me there is a 3rd option, so this is a mock up of that for a single notebook to get feedback before applying this to all! The idea here is that I sub-classes the nbconvert `ExecuteProcessor` to look for a cell-level metadata keyword that tells the processor to skip that cell. So the idea would be to follow the way `coordinates.ipynb` handles remote requests: to say in the text "if you have an internet connection, execute this, otherwise you can ignore the error that occurs." That notebook then provides hard-coded or cached values. With the new cell metadata key, we can tell our build machinery to skip cells that have remote requests when on CI.

cc @eteq


pllim:
So, this requires author to go into the raw text and add the incantation to every cell that needs to be skipped? What about a tutorial that use online query result for the whole thing (like cone search)?

adrn:
What do you mean by "raw text"? You can edit the cell-level metadata from the notebook GUI, but yes, that would be pretty annoying for the cone search tutorial! Will have to think about that..

PR #369, opened 2019-06-04
Hide filterTutorials from URL

### LINKED ISSUE #304: Hide the "q=filterTutorials" when showing all the tutorials
Right now to get all the tutorials, the URL has to end with the query `?q=filterTutorials`.  That's a bit ugly, and is an implementation detail leaking out into the user-space.  It also has the (admittedly minor) side-effect that if the text "featureTutorials" is in any notebook it gets highlighted. We should figure out how to have that not be necessary - I think this is probably just a javascript tweak to make the "no query string" default be equivalent to "filterTutorials".  But I haven't looked in detail, so @MananAgarwal might have more context/ideas.

adrn:
This PR is my attempt to fix #304.

This adjusts the way the search query is handled in javascript and assumes that no query string means filterTutorials.

### DIFF
--- a/tutorials/themes/tutorials-theme/layout.html
+++ b/tutorials/themes/tutorials-theme/layout.html
@@ -48,13 +48,6 @@
   <meta http-equiv='X-UA-Compatible' content='IE=edge,chrome=1'>
   <meta name='viewport' content='width=device-width, initial-scale=1.0, maximum-scale=1'>
   <meta name="apple-mobile-web-app-capable" content="yes">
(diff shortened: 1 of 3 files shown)


MananAgarwal:
This should hide filterTutorials.

"filterTutorials" infrastructure will be useful when we add Guides to Learn Astropy. Though we can use "filterGuides" for Guides and assume no query string to be "filterTutorials", but it will be inconsistent.

adrn:
OK, yea, point taken - we'll have to come up with another solution once we have guides.

PR #371, opened 2019-06-05
Create LICENSE

bmorris3:
It looks like the intended license (BSD) is written [here](https://github.com/astropy/astropy-tutorials/blob/master/metadata.cfg#L6) but I haven't found the file in the repo. Here's a license file!


adrn:
Thanks @bmorris3

PR #386, opened 2019-08-21
Fix UVES tutorial

adrn:
This will hopefully fix the issue in the UVES tutorial so that CircleCI starts building again.


adrn:
I squashed, but last test passed (hurray!) so I'm going to merge

pllim:
Thank you for tracking down the cert problem! I'll let you know if I hear updates from the server side of things. :grimacing:

PR #390, opened 2019-08-22
Mock remote HTTP requests

adrn:
This is a silly and somewhat crazy step towards dealing with our constantly failing circleCI builds that require firing off a bunch of remote HTTP requests (I'm looking at you, `vo_conesearch`). Here I've "mocked" a bunch of the astroquery and astropy classes and functions that make remote requests (which don't universally support caching) to instead point at localhost, and I've created a local http server to catch the requests and either (1) return results from a cache, or (2) pass the request through to the external site and then cache the results for the future. In principle, we could then add `~/.astropy/cache` along with this custom cache to be persistent data stores on circleCI (https://circleci.com/blog/persisting-data-in-workflows-when-to-use-caching-artifacts-and-workspaces/). 

This is a somewhat brittle system, and would require fiddling if/when the astroquery API changes, so I'm not sure we actually want to do this...

### DIFF
--- a/pip-requirements.txt
+++ b/pip-requirements.txt
@@ -13,3 +13,5 @@ dust_extinction>=0.7
 astro-gala
 dust_extinction>=0.7
 synphot
+flask
+requests
--- a/scripts/convert.py
+++ b/scripts/convert.py
@@ -11,6 +11,12 @@ from nbconvert.exporters import RSTExporter
 from nbconvert.writers import FilesWriter
 import nbformat
 
+import mocks  # just needs to be imported to run the code
+from mock_server import app, MOCK_SERVER_PORT
+
+# Start a Flask server to intercept and handle remote HTTP requests
+app.run(port=MOCK_SERVER_PORT)
+
 IPYTHON_VERSION = 4
 
 def clean_keyword(kw):
--- a/scripts/mock_server.py
+++ b/scripts/mock_server.py
@@ -1,0 +1,96 @@
+import hashlib
+import os
+import pickle
+from flask import Flask, request
+from urllib.parse import urlsplit, quote, unquote
+import requests
+
+
+MOCK_SERVER_PORT = 8001
+MOCK_SERVER = 'http://localhost:{port}'.format(port=MOCK_SERVER_PORT)
+
+app = Flask(__name__)
+
+this_path = os.path.split(os.path.abspath(__file__))[0]
+cache_path = os.path.abspath(os.path.join(this_path, '..', 'cache'))
+os.makedirs(cache_path, exist_ok=True)
+
+
+def pack_url(url):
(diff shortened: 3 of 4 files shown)


pllim:
This might be a can o' worms...

* `astroquery` already has some kind of mocking mechanism, though, admittedly `vo_conesearch` does not use it -- https://github.com/astropy/astroquery/blob/master/astroquery/utils/testing_tools.py
* Mocking does not guarantee a working end product for the user.
* There is a conflict of interest between these goal of testing/CI:
    * ensure the real code runs when user runs the tutorial, *or*
    * just want it to render successfully no matter what

As far as `vo_conesearch` is concerned (I cannot speak for the other queries), I can even throw in a fake "database" and "result" to be included as package data, which would avoid both mocking and network calls, but it wouldn't reflect on the real life use case. :woman_shrugging:

pllim:
p.s. @ceb8 could probably advise if there is a way to mock `astroquery.mast`.

adrn:
> astroquery already has some kind of mocking mechanism, though, admittedly vo_conesearch does not use it -- https://github.com/astropy/astroquery/blob/master/astroquery/utils/testing_tools.py

Right...

> Mocking does not guarantee a working end product for the user.

Maybe I didn't explain well enough how this works. This doesn't do what many mocking schemes do: the results returned by the mocks here will be a cached version of exactly what the non-mocked call would be.

> There is a conflict of interest between these goal of testing/CI: ensure the real code runs when user runs the tutorial, or just want it to render successfully no matter what

I don't follow. Currently, we are not running tests on CI (though, we should be) - we (or I) just want the tutorials site to build consistently and successfully on CI!

pllim:
What if you accidentally cached a bad call (like a 404 page)? How do you guarantee you are only caching the right stuff?

pllim:
Or do you mean you already manually built a good cache on the server?

pllim:
Also looks like your thingy timed out:
```
 * Serving Flask app "mock_server" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://127.0.0.1:8001/ (Press CTRL+C to quit)
make: *** [Makefile:43: html] Terminated
Too long with no output (exceeded 10m0s)
```

ceb8:
Yes astroquery requires mocked versions of all tests so that offline tests can be run. You could probably tie into that (look in astroquer/mast/tests/test_mast.py), but there are two problems:
1) Only a selection of responses are mocked, so you can't just monkey-patch a function with the mocked version and expect it to work
2) Because the mocked responses are fixed you are not testing that live service responses still work with the code, so you don't catch things like database changes or really anything that gets updated server size and breaks an example (in astroquery testing that's what the remote tests are for)

PR #392, opened 2019-08-23
Store and restore .astropy/cache

adrn:
This is yet another option for trying to improve build stability on circleCI with the tutorials that access remote services. This will not do anything for the `SkyCoord.from_name()` / Sesame calls, or, I believe, for vo_conesearch, but at least anything downloaded with `download_file()` should persist!

This obviates #350 and #390


eteq:
Coming in a bit late obviously, but I think this is good except for one minor suggestion: we should mention in the README or something that caching works this way, since it might confuse people.  But I can just make a PR on that.

PR #399, opened 2019-08-31
Clean up search result and summary text displayed on placards

### LINKED ISSUE #401: The term "filterTutorials" still showing in search field on "search.html" page
To follow up on #304. 

I have a feeling that this issue is indicative of a bug plaguing the Learn Astropy website in a way that warrants further investigation. 

Screenshot captured on September 10th, 2019 (HKT) on Chrome:

Screenshot 2019-09-10 01 06 48

It is interesting, because after you have searched for some search terms the field is showing the desired "Search the universe of Astropy" as coded in the "navbar.html" instead. See say if we changed the URL to "http://learn.astropy.org/search.html?q=scipy.integrate#" and then delete the original search term from the search field:

Screenshot 2019-09-10 01 11 10

adrn:
This is the first of a few PRs I'll submit to improve the theme for the Learn site. This first one is focused on improving the summary text that is displayed in the search results, or when no search is performed and all tutorials are shown. 

Fixes #401


adrn:
This now makes some other improvements that should fix #401. Specifically, I split up the concept of the search target and the search query. Now, the query string really is just the query string (i.e. no magic "filterTutorials" to specify that we want to search the tutorials), and there is an additional HTML parameter "target" that, by default, is set to `target="tutorials"`, but in the future we can use this parameter to support searching in other contexts like examples, guides, etc. or all of them together.

adrn:
(so it would be great to get your eyes on this one more time @MananAgarwal if you have some time!)

tutorials/themes/tutorials-theme/static/searchtools.js
```diff
@@ -773,86 +820,135 @@ var Search = {
...
+                                  parBreakIdx - summaryTextIdx);
+        excerpt = cleanupRSTString(text.substr(summaryTextIdx, summaryLen));
+        excerpt = excerpt + '...';
+
+      } else {  // No summary header found
+        excerpt = 'No tutorial summary found.';
+      }
+    } else {
+      $.each(keywords, function(i, x) {
+        var keyIdx = textLower.indexOf(hlwords[i].toLowerCase(), startIdx);
```
MananAgarwal: When we use the "filters" to search for the results, "hlwords" will look like "filterKEYWORD". Hence, indexof won't find any instance of the hlword.

Screenshot 2019-09-16 at 2 35 08 PM

We want to first check if the hlword has "filter" in it. If yes then we have to first strip it, else we can directly use it. Something similar to want we are doing in lines 942 - 950 (i.e. in $.each(hlwords, function())

tutorials/themes/tutorials-theme/static/searchtools.js
```diff
@@ -773,86 +820,135 @@ var Search = {
...
+      newStr = newStr.replace(/(?:www|https?)[^\s]+/gm, "");
+
+      // Remove RST formatting:
+      newStr = newStr.replace(/(`)/gm, "");
+
+      return newStr;
+    }
+
+    // Start looking at search text below the main header
+    var startIdx = textLower.indexOf('====\n', 0) + 5;
```
MananAgarwal: Also, I think we should update startIdx so that we retrieve more informatic text about the tutorial. 

Screenshot 2019-09-16 at 2 44 22 PM

adrn:
This is obviated by the website re-work in progress...

embray:
What is the "website re-work in progress"?  Is that still actually in progress?  This PR was over a year ago and the issue it was addressing still isn't fixed.

PR #411, opened 2019-11-01
Removing broken link from README

bmorris3:
The link to http://www.astropy.org/astropy-tutorials/#for-contributors no longer works, simply removing it from the README. The subsequent link on the page still works and serves the same purpose.


adrn:
Hmm, the link works for me! What behavior are you seeing?

kakirastern:
The link works for me too... its expected behavior is to "point" to the `#for-contributors` tag of the http://learn.astropy.org/ home page through redirection from  http://www.astropy.org/astropy-tutorials/

PR #425, opened 2020-01-21
Fix Giles URL

pllim:
cc @Cadair


adrn:
So, the old URL is deprecated? Sorry, I just don't know the context here!

Cadair:
yes, it should (now) be still working as I put it back, but I would like to shut it down eventually.

PR #495, opened 2021-09-21
Switch to using nbcollection + github actions

adrn:
Demonstration of switching the execute & convert machinery to nbcollection with github actions instead of custom scripts + circleCI.


adrn:
Oops, I think I need to first add a workflow in `main` to get it to see the new workflow...

adrn:
Note: there are real failures in the conesearch tutorial because of an issue with the VO server (comments by @pllim in slack). In the long run, we still need to implement a more robust solution for handling remote data. One idea would be to distribute the `~/.astropy/cache` needed by the tutorials with this repo so that CI doesn't do *any* remote data access. In the short term, we might want to just skip the conesearch tutorial in the build process...

adrn:
@jonathansick I think this is now a working proof of concept! Here is the resulting artifact / zip file of the rendered tutorials: https://github.com/astropy/astropy-tutorials/suites/3848792615/artifacts/95489955
Should we merge this and have you add the cross-deploy in a separate PR?

adrn:
(Tried to get the circleci error to go away, because I removed the webhook, but no luck...)

jonathansick:
@adrn This looks great. Since this doesn't affect the current production (afaik), this should be safe to merge now and I'll contribute the cross-deploy into the astropy/learn-astropy repo.

I _might_ need to adjust some of the CSS selectors in astropy/learn-astropy-librarian in order to gather the new notebooks since they aren't wrapped in the Sphinx HTML structure anymore. That shouldn't be an issue, though.

adrn:
Great! OK I'll do it then - #yolo

PR #497, opened 2021-09-27
Add repository-dispatch workflow to deployment

jonathansick:
This PR adds a "repository dispatch" step to the build workflow. What this does is trigger a GitHub Actions run in the astropy/learn-astropy repository. The payload for the trigger includes a reference to the workflow artifact, which will allow the learn-astropy build to download the tutorials artifact.

This requires a GitHub Secret to be set up on the tutorials repo named `DISPATCH_GITHUB_TOKEN` that has write access to the astropy/learn-astropy repository. See https://github.com/peter-evans/repository-dispatch#token

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,21 @@ jobs:
         run: |
           nbcollection convert --flatten --build-path=. -v --make-index --index-template=templates/index.tpl tutorials --exclude=conesearch  # TODO: remove this exclude!
 
+      - name: Name artifact
+        id: nameartifact
+        run: |
+          echo "::set-output name=artifactName::rendered-tutorials-${{ github.sha }}"
+
       - uses: actions/upload-artifact@v2
         with:
-          name: rendered-tutorials-${{ github.sha }}
+          name: ${{ steps.nameartifact.outputs.artifactName }}
           path: _build
+
+      - name: Dispatch Learn Astropy deployment
+        uses: peter-evans/repository-dispatch@v1
+        if: ${{ (github.event_name == 'push') && (github.ref == 'refs/heads/main') }}
+        with:
+          token: ${{ secrets.DISPATCH_GITHUB_TOKEN }}
+          repository: astropy/learn-astropy
+          event-type: tutorials-build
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -39,7 +39,21 @@ jobs:
...
+          echo "::set-output name=artifactName::rendered-tutorials-${{ github.sha }}"
+
       - uses: actions/upload-artifact@v2
         with:
-          name: rendered-tutorials-${{ github.sha }}
+          name: ${{ steps.nameartifact.outputs.artifactName }}
           path: _build
+
+      - name: Dispatch Learn Astropy deployment
+        uses: peter-evans/repository-dispatch@v1
```
pllim, line 53: Unless you completely trust this repo, it is safer to pin at the SHA that corresponds to the tag.

jonathansick:
@adrn Would you mind taking a look at this one? I don't think I can assign the review to you because of repo access.

Keep in mind that you'll also need to set up the GitHub Personal Access Token and add it as a secret. The token could come from a user account or a custom GitHub App installed into astropy/learn-astropy.

adrn:
> I don't think I can assign the review to you because of repo access.
Whoops! I just added you to the Learn Astropy Maintainers team - that was an oversight on my part...

> Keep in mind that you'll also need to set up the GitHub Personal Access Token and add it as a secret. The token could come from a user account or a custom GitHub App installed into astropy/learn-astropy.
As in, when we merge this, right?

adrn (approved):
Cool, I think this looks good! So the DISPATCH_GITHUB_TOKEN should be a repo access token? With what privileges?

adrn:
BTW: You might have permissions now to set up the token yourself?

jonathansick:
Awesome, I added the secret to the repo. Right now I'm using a Personal Access Token, but unfortunately, I can't scope it to a specific repository. I'll make a to-do for myself to investigate writing a new action that runs through a custom GitHub App because that apps essentially create tokens that are scoped to specific orgs/repos.

pllim:
PAT is probably unavoidable. If necessary, another maintainer can refresh it in admin page.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#granting-additional-permissions

PR #502, opened 2021-11-20
Fix several broken links

keflavich:
There were several links that were broken because line breaks were put in the middle of the URL.

The Jupyter style guidelines link was dead. I put in the first google hit for jupyter style guidelines; I'm not certain it's the appropriate link.


keflavich:
The fact that this has gone unnoticed for 4 years suggests there hasn't been much visibility for this document.  It may warrant further review now that it is more prominently featured on the frontpage:

image

Also it looks like the problem was introduced probably by an automated `.md`->`.rst` converter that truncated lines at 80 characters.

adrn:
Thanks for catching this! We actually just switched over the web front-end and restructured this repo, which is probably why the links are broken. It's on our (@jonathansick) TODO list to revamp the contributor guide, but this helps - thanks!

PR #507, opened 2021-11-30
Enabled ExtractOutputPreprocessor

jonathansick:
This preprocessor moves figures out of the HTML output. They should still upload normally to GitHub Pages though because we capture and upload the build artifact as a whole directory rather than file-by-file.


adrn:
See #530 for an alternate solution with the new makefile build!

Issue #506, opened 2021-11-29
BUG in tutorials/UVES/UVES.ipynb function read_setup()


jzoubian:
During astroinfo 2021, we used your great tutorials and one of the participant founds a bug in the tutorials UVES, in the function read_setup()

```
def read_setup(filename):
    '''Get setup for UVES spectrum from the ESO pipeline

    Parameters
    ----------
    filename : string
    name of the fits file with the data

    Returns
    -------
    exposure_time : float
    wavelength_zero_point : float
    optical_arm : string
    '''
    sp = fits.open(filelist[0])
    header = sp[0].header

    return header['EXPTIME'], header['CRVAL1'], header['HIERARCH ESO INS PATH'
```

`sp = fits.open(filelist[0])` should be `sp = fits.open(filename)`.

In the loop :
```
for f in filelist:
    print(read_setup(f))
```
all the setup are the same because it is always the same print filelist[0].

adrn:
Thanks for catching this! I implemented a fix in #506

PR #516, opened 2022-01-04
Fixes read_setup() bug reported in #506

adrn:
See #506 for the bug and quick fix, implemented here.


adrn:
Failure unrelated...

PR #523, opened 2022-01-05
Only run modified notebooks in PRs, with an optional label override

### LINKED ISSUE #519: In PRs, only execute notebooks that have been modified
We have discussed this idea for a while, but we should implement functionality such that in a pull request, nbcollection only executes notebooks that have been modified. This will significantly reduce the CI headaches we have in PRs! If we do this, we should add an optional label that would trigger executing all notebooks, and add a weekly cron job that executes all notebooks.

This could either be implemented here with bash scripts that use git commands to get the list of modified files and only passes those names to nbcollection for execution. Or, we could add support for this in nbcollection itself.

adrn:
Refreshing some of the CI config files so that, by default, PR builds only run notebooks that have been modified.

Note: This currently contains a test commit that modifies two notebooks, as a test of the new github actions workflow. That commit should be removed before merging, if this works!

Fixes #519

### DIFF
--- a/.github/get_modified_tutorials.py
+++ b/.github/get_modified_tutorials.py
@@ -1,0 +1,24 @@
+import sys
+from git import Repo
+
+
+def main(repo_path):
+    r = Repo(repo_path)
+
+    # NOTE: assumes the main branch is named "main"
+    files_changed = r.git.diff(
+        f'{str(r.active_branch)}..main',
+        '--name-only').split("\n")
+    files_changed = [f for f in files_changed if f.endswith('.ipynb')]
+    print(" ".join(files_changed))
+
+
+if __name__ == "__main__":
+    try:
+        repo_path = sys.argv[1]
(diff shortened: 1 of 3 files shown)


adrn:
Oo it worked!

adrn:
And it looks like the label worked as well! Changing this PR title - I think this can be merged.

adrn:
Now it works even if there are no modified tutorials - I think this is ready to merge!

jonathansick (approved):
Nice, simple solution. I like separating the deployment and testing workflows. Only thing I might want to add is having an artifact of the build HTML uploaded to assist with reviews; but we can handle that later.

adrn:
> Only thing I might want to add is having an artifact of the build HTML uploaded to assist with reviews; but we can handle that later.

Yea, that would be awesome to have! I'll look into some options for solutions there.

PR #550, opened 2022-04-12
New tutorial to calculate radial velocity from multiple measurements

alexbinks:
This tutorial is designed to assess the multiplicity status of a
star and calculate a "final" radial velocity for stars when there
are several (> 2) individual measurements. Whilst intended for
RV measurements, the method can be applied for any
parameter that needs to be weighted from different sources.
Principally the program:
1) reads in a table of RV measurements for any set of stars
2) makes a decision on whether the star is multiple or likely single
3) calculates a final RV from the individual measurements.

- [ ] Confirm that any contributed tutorials contain a complete Introduction which includes an Author list, Learning Goals, Keywords, Companion Content (if applicable), and a Summary.


hamogu:
Thanks for contributing! This looks interesting, but I think you might want to see if the indexing logic can be simplified a bit. At least I don't understand it - but maybe that's just because I don't know that much about the topic.

kelle:
This looks really promising! Thanks so much for the contribution and we'll work on giving it a robust review ASAP. I'm also going to start a thread in the #learn channel on slack.

ceb8:
Hello @alexbinks! 

I realise it has now been a number of years since you submitted this and that you have probably moved on to new projects. However, I think this is a useful notebook and would like to see it added to Learn. To that end I am happy to do the editing to make it conform to our coding standards. I just have two questions about the code itself, and need you to say that you are happy for me to do this (I will link you to the notebook before it goes live so that you get a chance to verify it before we make it public).

My questions are:

1. In the `test_mult` function you check where the calculated $\Delta RV$ values are greater than the max as:  `ee = c > d_max`,  `x = np.array(np.where(ee))[0]`. Why do you only use the first row of `np.array(np.where(ee))`?

2.  One line later you remove the unphysical values with `rv = np.delete(trv, x[0], 0)`. Why do you only use the first index of `x`?

I apologise for how long we've let this sit. Let me know what you think of my proposed plan of action.

Thank you for your contribution!

hamogu:
I'll have a look, I'm one of the co-authors after all. I won't get to that until next week though.

Issue #532, opened 2022-01-18
Explore possibilities for linking the executed (with output) and rendered (HTML) notebooks in a PR


adrn:
The build artifacts from a PR's CI run should contain these things. Is there a resource we can use that will take the artifact, unzip it, and serve it at a url?

adrn:
@Cadair: We used to use Giles to provide a link to our built Sphinx site in a PR. We've switched to using nbconvert more directly, but it would be cool to link to the converted/rendered HTML notebooks in a PR. Do you know of relevant infrastructure we could use to do that?

Cadair:
Well if you use circleci and upload the html from nbconvert to circleci's artifacts then you could still use giles as you can configure it to link to any file on circleci's aritfacts.

I am not aware of any other CI provider other than circle CI which lets you direct link to HTML pages and be able to see them in the browser without any issues, so that's why I say use circleci.

jonathansick:
Maybe we could set up a staging server? Deploy into this repo's GitHub Pages with a directory prefixed by the PR number? We could even have a GitHub Actions workflow that cleans up the staged directory in gh-pages when a PR is closed.

There is the security concern of hosting content from PRs (though it wouldn't be highly visible and wouldn't be indexed by search engines, so it may be an acceptable risk)

Cadair:
You *could* but given the circle ci provides that feature for free why would you?

PR #554, opened 2022-04-14
Upload artifacts in PR build

eteq:
- [x] Confirm that any contributed tutorials contain a complete Introduction which includes an Author list, Learning Goals, Keywords, Companion Content (if applicable), and a Summary.

This PR adds the upload artifact stage to the PR workflow (but not the actual deploy, of course), so that we can see the rendered version of a notebook before it's merged. This could be thought of as a first step of #532 (although that needs more before I think we would consider it closed - see the discussion there).

Note the second commit here should be removed before merging - that's just to trigger the actual build process on one of the notebooks.  So the intent is to back out that last commit before merging (although it leaves an unnecessary comment).


.github/workflows/prs.yml
```diff
@@ -46,3 +46,13 @@ jobs:
...
+
+      - name: Name artifact
+        id: nameartifact
+        run: |
+          echo "::set-output name=artifactName::rendered-tutorials"
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: ${{ steps.nameartifact.outputs.artifactName }}
+          path: _build
```
adrn, line 58:
```suggestion
          path: _build

```

adrn:
@eteq This is fine with me! Can you remove the noop commit and then we can merge?

adrn (approved):
I removed the commit for ya - thanks!

PR #602, opened 2024-09-24
Update versions like a human dependabot

adrn:
This is just to deal with some out of date GitHub actions workflows. The build will probably still fail but @jeffjennings and I are working on getting the tutorials build going again!


kelle (approved):
Amazing!

PR #606, opened 2024-09-26
[WIP] Use individual notebook requirements

jeffjennings:
Removes the top-level `requirements.txt` in favor of individual `requirements.txt` specified in each folder in `tutorials`.

- [x] Confirm that any contributed tutorials contain a complete Introduction which includes an Author list, Learning Goals, Keywords, Companion Content (if applicable), and a Summary.

### DIFF
--- a/tutorials/astropy-coordinates/requirements.txt
+++ b/tutorials/astropy-coordinates/requirements.txt
@@ -1,0 +1,4 @@
+astropy
+astroquery
+matplotlib
+numpy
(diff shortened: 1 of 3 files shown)


adrn (commented):
Overall, I think this will be a good change for users! 

I'd like to discuss one implication of this change, which is that in this new way of managing requirements, pinning dependencies now requires updating many files. It would be nice if we didn't have to keep track of this. But I can't think of a better way of managing that! Thoughts? 

Relatedly, I can see a use case (at least for us as developers) for wanting to be able to install all dependencies in one go. The old top-level requirements.txt file provided this. Would it make sense to provide a makefile command that just loops through all of the tutorial directories and installs all requirements?

jeffjennings:
Yes it's a good idea! Unfortunately there are a few cases where the version for the same requirement differs, making it hard to have a single, top-level set of requirements.

